PR #9, opened 2025-10-17
[Github] Remove extra github utilities/files/workflows

jonathonpenix:
Broadly, I don't think we want any of these right now or, in the case of CODEOWNERS, we probably want something like that eventually, but we don't have specific POCs dedicated right now or expect significant contributions.

So, just remove these for now.


apazos:
LGTM

navaneethshan (approved):

Yethendra (approved):

PR #10, opened 2025-10-17
[Github] Update pre-checkin to work with fork -> PR model

jonathonpenix:
Our pre-checkin.yml seemed targeted at PRs coming from branches in the same repository, and doing a fork -> PR resulted in build failures. Try to address that.

Additionally, we don't expect musl-embedded to share the cpullvm precheckin anymore, so we can strip out that code too to make this change simpler.

### DIFF
--- a/.github/workflows/pre-checkin.yml
+++ b/.github/workflows/pre-checkin.yml
@@ -1,15 +1,6 @@
 name: pre-checkin
 
-on:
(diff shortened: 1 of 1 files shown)


navaneethshan (approved):

jonathonpenix:
The manual squash not such a great idea--will merge once the build completes.

PR #19, opened 2025-10-20
Convert files from Windows line endings to Unix

jonathonpenix:
I'm not sure why these had Windows line endings in the first place, but I'm reformatting these to keep this consistent with the rest of LLVM/our other projects and hopefully prevent any surprises when looking at git diff output (which highlights the Windows line endings apparently).


navaneethshan (approved):
Yes, these files should have Unix line endings, it must have got Windows line endings as I used Visual Studio as editor from Windows, I converted it to Unix, looks like I overrode it at somepoint.

PR #23, opened 2025-10-22
[Github] Prevent workflows from running on forks

jonathonpenix:
Right now, our workflows also seem to run in my fork (without any additional configuration, AFAIK). This ends up being a bit annoying as the qcom-preflight-checks run on pushes and fails unless "Dependency graph" is explicitly enabled. Generally, I don't think people will be configuring runners, etc. in their forks either, so just disable all the workflows outside the main repo.

### DIFF
--- a/.github/workflows/pre-checkin.yml
+++ b/.github/workflows/pre-checkin.yml
@@ -5,6 +5,7 @@ on:
 
 jobs:
   base:
+    if: github.repository == 'qualcomm/cpullvm-toolchain'
     uses: ./.github/workflows/base-build.yml
     with:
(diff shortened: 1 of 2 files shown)


jonathonpenix:
Gentle ping, these still run in forks.

jonathonpenix:
Closing this--https://github.com/qualcomm/cpullvm-toolchain/pull/83 is a reset for the workflows, and will just keep this in mind as we go (and, has the one workflow already guarded)

jonathonpenix:
I take that back--this is still an issue. This has been sitting for a few months and should be an unobtrusive change, so merging https://github.com/qualcomm/cpullvm-toolchain/pull/95

PR #26, opened 2025-10-24
Updated project overview

apazos:


jonathonpenix (approved):
Build failure should be that we don't have https://github.com/qualcomm/cpullvm-toolchain/commit/b025608e34fd64a7fa9a6b435285d0e8f73cbbe8 in the branch--I'll do that separately (and the readmes shouldn't impact the build at all so I'm not worried about that)

I'm fine with merging and confirming that the links are working as expected then

PR #27, opened 2025-11-01
[Github] Update pre-checkin to work with fork -> PR model (#10)

jonathonpenix:
Our pre-checkin.yml seemed targeted at PRs coming from branches in the same repository, and doing a fork -> PR resulted in build failures. Try to address that.

Additionally, we don't expect musl-embedded to share the cpullvm precheckin anymore, so we can strip out that code too to make this change simpler.

This is basically a cherry-pick of what we did in main--I'm bringing this into the release branch just in case we need to make any additional changes.

### DIFF
--- a/.github/workflows/pre-checkin.yml
+++ b/.github/workflows/pre-checkin.yml
@@ -1,15 +1,6 @@
 name: pre-checkin
 
-on:
-  pull_request:
-  workflow_call:
-    inputs:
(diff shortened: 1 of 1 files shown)


jonathonpenix:
This is obsolete after https://github.com/qualcomm/cpullvm-toolchain/pull/33, closing

PR #33, opened 2025-11-09
Workflow changes for aarch64 builds

navaneethshan:


qualcomm-software/embedded/scripts/build_aarch64.sh
```diff
@@ -0,0 +1,420 @@
...
+SKIP_TESTS="false"
+JOBS="${JOBS:-$(nproc)}"
+
+log() { echo -e "\033[1;34m[precheckin]\033[0m $*"; }
+warn() { echo -e "\033[1;33m[warn]\033[0m $*"; }
+
+usage() {
+  cat <<'EOF'
+Usage:
+  build.sh [options]
```
navaneethshan: build_aarch64.build

qualcomm-software/embedded/scripts/build_aarch64.sh
```diff
@@ -0,0 +1,420 @@
...
+
+Options:
+  --artifact-dir <path>       Directory to copy final tarball
+  --skip-tests                Skip LLVM test steps
+  --aarch64-sysroot <path>    AArch64 sysroot (default: /usr/aarch64-linux-gnu)
+  --clean                     Delete and recreate build/install dirs
+
+Examples:
+  ./build.sh --artifact-dir /tmp/artifacts
```
navaneethshan: same

qualcomm-software/embedded/scripts/build_aarch64.sh
```diff
@@ -0,0 +1,419 @@
...
+# The script automates cloning, patching, building LLVM, ELD linker and musl-embedded,
+# and packaging artifacts for ARM and AArch64 targets.
+
+# Note: Ensure that `ELD_BRANCH` and `MUSL_EMBEDDED_BRANCH` match the current repository branch
+# to maintain consistency across all dependencies.
+
+set -euo pipefail
+
+readonly ELD_REPO_URL="https://github.com/qualcomm/eld.git"
+readonly ELD_BRANCH="main"
```
jonathonpenix: It doesn't matter right now (and if we keep the separate script), but since we decided earlier that we were going to have this mirror the appropriate branch, we might as well change this to match I think

qualcomm-software/embedded/scripts/build_aarch64.sh
```diff
@@ -0,0 +1,419 @@
...
+ARM32_BM_TRIPLE="arm-none-eabi"
+AARCH64_BM_TRIPLE="aarch64-none-elf"
+COMPILER_RT_ARM32_BM_BUILDDIR="${WORKSPACE}/build/compiler-rt/arm32/baremetal"
+COMPILER_RT_AARCH64_BM_BUILDDIR="${WORKSPACE}/build/compiler-rt/aarch64/baremetal"
+COMPILER_RT_ARM32_BM_FLAGS="--target=arm-none-eabi -mcpu=cortex-a9 -ffunction-sections -fdata-sections -mfloat-abi=softfp -mfpu=neon -nostdlibinc"
+COMPILER_RT_AARCH64_BM_FLAGS="--target=aarch64-none-elf -mcpu=cortex-a53 -ffunction-sections -fdata-sections -nostdlibinc"
+
+# aarch64 cross compilation environment
+SYS_ROOT_AARCH64="/usr"
+AARCH64_LIBC="$SYS_ROOT_AARCH64/aarch64-none-linux-gnu/libc"
```
jonathonpenix: What is this/where is this sourced from? This doesn't look like anything standard that people will have installed (either by default or with anything we've documented in the README)

qualcomm-software/embedded/scripts/build_aarch64.sh
```diff
@@ -0,0 +1,419 @@
...
+AARCH64_LINUX_FLAGS="--target=aarch64-linux-gnu --sysroot=$AARCH64_LIBC --gcc-toolchain=$SYS_ROOT_AARCH64 -pthread"
```
jonathonpenix: I really don't think we should be linking with GNU tools here--I realize we're linking with GNU ld in the x86 Linux builds today, but I'd rather see us switch everything over to lld/the llvm tools rather than bringing in GNU toolchains for cross compiling here.

qualcomm-software/embedded/scripts/build_aarch64.sh
```diff
@@ -0,0 +1,419 @@
...
+
+#--- Help the cross-linker find libdl/librt in the AArch64 sysroot ---
+RPATH_LINK_FLAGS="-Wl,-rpath-link,$AARCH64_LIBC/lib -Wl,-rpath-link,$AARCH64_LIBC/usr/lib \
+                  -Wl,-rpath-link,$AARCH64_LIBC/lib64 -Wl,-rpath-link,$AARCH64_LIBC/usr/lib64 \
+                  -L$AARCH64_LIBC/lib -L$AARCH64_LIBC/usr/lib -L$AARCH64_LIBC/lib64 -L$AARCH64_LIBC/usr/lib64"
```
jonathonpenix: Can we just pass in sysroot to the compiler driver instead? My expectation would be that the right paths should end up in the linker invocation

qualcomm-software/embedded/scripts/build_aarch64.sh
```diff
@@ -0,0 +1,419 @@
...
+  -DLLVM_EXTERNAL_PROJECTS="eld" \
+  -DLLVM_EXTERNAL_ELD_SOURCE_DIR="llvm/tools/eld" \
+  -DLLVM_DEFAULT_TARGET_TRIPLE="aarch64-unknown-linux-gnueabi" \
+  -DLLVM_TARGET_ARCH="arm-linux-gnueabi" \
+  -DLLVM_BUILD_RUNTIME="OFF" \
+  -DLIBCLANG_BUILD_STATIC="ON" -DLLVM_POLLY_LINK_INTO_TOOLS="ON" \
+  -DCMAKE_C_COMPILER="clang" -DCMAKE_CXX_COMPILER="clang++" \
+  -DCMAKE_CXX_FLAGS="-stdlib=libc++" \
+  -DLLVM_ENABLE_ASSERTIONS:BOOL="${ASSERTION_MODE}" \
+  -DLLVM_ENABLE_PROJECTS="llvm;clang;polly;lld;mlir" \
```
jonathonpenix: If we keep this, we should probably be pruning down the projects we build to what we actually use/need if only to keep build times somewhat reasonable.

qualcomm-software/embedded/scripts/build_aarch64.sh
```diff
@@ -0,0 +1,419 @@
...
+ninja -C "${BUILD_DIR}/llvm"
+
+log "Installing LLVM native X86 [Stage 1]"
+ninja -C "${BUILD_DIR}/llvm" install
+
+# ---  Stage 2: host = AArch64, targets = ARM;AArch64 ---
+log "[Stage 2] Configuring Cross-compiling LLVM for AArch64..."
+mkdir -p "$BUILD_DIR_AARCH64" "$INSTALL_DIR_AARCH64"
+
+cmake -G Ninja \
```
jonathonpenix: As a more general comment, did we check https://llvm.org/docs/HowToCrossCompileLLVM.html? This was updated relatively recently (https://github.com/llvm/llvm-project/commit/ba11e1e5222cb5bdcd8d63f23fb994cc0a8ffb3c) and I'd much prefer we align towards that if possible (or, at least that we have a good reason _not_ to if we aren't)

jonathonpenix: Another note here is that, IIUC, this is going to be building with libstdc++ (whereas x86 is currently libc++). We've talked about changing that on the x86 side, but is this intended?

If nothing else, this is probably another dependency people need to be aware of when obtaining a sysroot. Even besides that though, it seems like we should probably be consistent between platforms where possible.

Edit: and just to be a bit more clear, I'm guessing libstdc++ is going to be more expedient than building/sourcing a cross libc++ so I don't have an issue with it as long as this is expected/we are using a reasonably up-to-date libstdc++.

qualcomm-software/embedded/scripts/build_aarch64.sh
```diff
@@ -0,0 +1,419 @@
...
+    ninja install
+    popd >/dev/null
+echo "c++ libs install ..."
+done
+echo "Build and installation complete."
+
+# --- Create artifact ---
+log "Creating artifact tarball"
+pushd "${INSTALL_DIR}" >/dev/null
+cp -r aarch64/bin aarch64/lib aarch64/libexec aarch64/share aarch64/tools .
```
jonathonpenix: Am I understanding correctly that this is taking a subset of the toolchain we cross compiled and mixing it in with the native x86 one?

If so, I can't point to a specific thing that will break offhand as-is (I don't know how much of ex: include would be target-specific), but this doesn't seem like a good idea to me

Seems like we probably should be building out the aarch64 sysroot with the libs as we go instead (assuming we are building the libs, see top level comment) and using just that resulting toolchain.

jonathonpenix (changes requested):
As an initial higher-level comment/question on this: do we have any expectation that we maintain/do any development on our 21.x branch?

With this, we have 4 nearly identical build recipes we'll be maintaining in parallel (not counting whatever happens with Windows) which can't be a good thing. Looking at this and assuming we have at least some intention of maintaining these configs, can we discuss either:
* merging these, even if it is just something quick (putting the native-aarch64 stuff under some flag in the build.sh and having the precheckins invoke that directly)
* or, since we've discussed copying runtimes/distributing those as a separate package for Windows already, at least removing all the runtime building and just do a minimal build here?

qualcomm-software/embedded/scripts/build_aarch64.sh
```diff
@@ -0,0 +1,419 @@
+#!/usr/bin/env bash
```
jonathonpenix: If we keep this, please also make this file executable

Yethendra:
> As an initial higher-level comment/question on this: do we have any expectation that we maintain/do any development on our 21.x branch?

Do you mean we can have a single script for both x86 and aarch64?

1 - Stage#1 - linux x86 
2 - Stage#2 - linux aarch64 

If we agree the above, we can have a single script which generated toolchain for x86 and aarchh64 which we can package as two different toolchains.

Yethendra:
- We agreed that detailed work is needed to de-duplicate and create a single common script for all variants.
- The goal of this recipe/PR is to provide the toolchain and assets for upload to unblock customers. All scripts and workflows will be updated once your final work is complete.
- Therefore, I recommend not merging this PR. Instead, use its artifacts for now, and we can upload the finalized workflow after the de-duplication work is finished.

qualcomm-software/embedded/scripts/build_aarch64.sh
jonathonpenix: Just to give a more explicit example of what I'm wondering about, should we be setting ex: `LLVM_HOST_TRIPLE` or is there a particular reason we don't need to?

Again I don't know if there's actually an issue there, but based on how it is documented in the above and in some places in llvm (ex: lldb calls out not setting this as possibly causing issues), I'm wondering if we're going to be getting into issues.

Looking at my cache file when running `build_aarch64.sh` locally, I see `LLVM_HOST_TRIPLE:STRING=x86_64-unknown-linux-gnu` so I don't think we're inferring the triple we presumably want.

Yethendra: AFAIK, the LLVM will auto-detect the host triple based on the build machine’s OS / architecture. In our case its Linux/x86 as host, generating code for ARM/ARM64

jonathonpenix: My understanding is `LLVM_HOST_TRIPLE` is about the host the built LLVM will run on (aarch64 Linux for us) rather than the host it is built on (x86 Linux) (so, seeing `LLVM_HOST_TRIPLE:STRING=x86_64-unknown-linux-gnu` seems odd, unless I'm misunderstanding)

qualcomm-software/embedded/scripts/build_aarch64.sh
jonathonpenix: TODO: we're going to add a reference to this in the README in a subsequent patch

qualcomm-software/embedded/scripts/build_aarch64.sh
jonathonpenix: Keeping this as GNU ld for now, will revisit this in mainline

qualcomm-software/embedded/scripts/build_aarch64.sh
jonathonpenix: LLVM_HOST_TRIPLE to aarch64, to investigate passing `--sysroot` as part of link flags as well rather than explicitly passing `-L/-rpath-link` (can maybe set via CMAKE_SYSROOT, see https://llvm.org/docs/HowToCrossCompileLLVM.html)

jonathonpenix: Leave this as-is (libstdc++) for now, but this is something we can revisit in the future (in mainline).

navaneethshan: I think this is taken care, please chekc.

.github/workflows/base-build.yml
```diff
@@ -0,0 +1,43 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: self-hosted
+
+    steps:
+      - name: Cleanup workspace (safe)
```
jonathonpenix: Why is this needed? From past discussions, I thought each workflow run was it's own fresh VM (or, something like that) and my impression is/was that the dir `GITHUB_WORKSPACE` points at should be empty at start

.github/workflows/nightly-aarch64.yml
```diff
@@ -0,0 +1,11 @@
...
+
+on:
+  schedule:
+    - cron: '0 5 * * *' # 9 PM PT (UTC)
+
+jobs:
+  base:
+    uses: ./.github/workflows/base-build.yml
+    with:
+      build-script: "src/qualcomm-software/embedded/scripts/build.sh --skip-clones --aarch64-build"
```
jonathonpenix: Should `--skip-clones` be deleted?

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -42,6 +44,7 @@ Options:
   --artifact-dir <path>       Directory to copy final tarball
   --skip-tests                Skip LLVM test steps
   --aarch64-sysroot <path>    AArch64 sysroot (default: /usr/aarch64-linux-gnu)
+  --aarch64-build             Aarch64 build
```
jonathonpenix:
```suggestion
  --aarch64-build             AArch64 build
```

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -79,10 +84,15 @@ COMPILER_RT_AARCH64_BM_BUILDDIR="${WORKSPACE}/build/compiler-rt/aarch64/baremeta
 COMPILER_RT_ARM32_BM_FLAGS="--target=arm-none-eabi -mcpu=cortex-a9 -ffunction-sections -fdata-sections -mfloat-abi=softfp -mfpu=neon -nostdlibinc"
 COMPILER_RT_AARCH64_BM_FLAGS="--target=aarch64-none-elf -mcpu=cortex-a53 -ffunction-sections -fdata-sections -nostdlibinc"
 
-# --- Prepare build/install dirs ---
+# aarch64 cross compilation environment
+SYS_ROOT_AARCH64="/usr"
+AARCH64_LIBC="${SYS_ROOT_AARCH64}/aarch64-none-linux-gnu/libc"
```
jonathonpenix: We briefly discussed this earlier, but as far as I can tell none of the packages we've documented (in the READMEs or internally) actually result in a `/usr/aarch64-none-linux-gnu` folder. Are we sourcing this from somewhere else?

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -79,10 +84,15 @@ COMPILER_RT_AARCH64_BM_BUILDDIR="${WORKSPACE}/build/compiler-rt/aarch64/baremeta
...
+AARCH64_LINUX_FLAGS="--target=aarch64-linux-gnu --sysroot=${AARCH64_LIBC} --gcc-toolchain=${SYS_ROOT_AARCH64} -pthread"
```
jonathonpenix: This looks unused

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -138,7 +148,47 @@ log "Installing LLVM"
...
+    -DLLVM_BUILD_RUNTIME="OFF" \
+    -DLIBCLANG_BUILD_STATIC="ON" \
+    -DLLVM_POLLY_LINK_INTO_TOOLS="ON" \
+    -DCMAKE_SYSTEM_NAME="Linux" \
+    -DCMAKE_SYSTEM_PROCESSOR="aarch64" \
+    -DCMAKE_SYSROOT="${AARCH64_LIBC}" \
+    -DCMAKE_C_COMPILER="clang" \
+    -DCMAKE_CXX_COMPILER="clang++" \
+    -DCMAKE_C_FLAGS="--target=aarch64-linux-gnu \
+                     --sysroot=${AARCH64_LIBC} \
```
jonathonpenix: I'm guessing we can remove `--sysroot` here (and in the `CMAKE_CXX_FLAGS` below) since we set it in `CMAKE_SYSROOT` above

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -344,26 +392,32 @@ for VARIANT in "aarch64-none-elf" "aarch64-pacret-b-key-bti-none-elf" "armv7-non
...
 # --- Create artifact ---
 log "Creating artifact tarball"
 pushd "${INSTALL_DIR}" >/dev/null
 short_sha="$(git -C "${SRC_DIR}" rev-parse --short HEAD)"
-tar_file="${ELD_BRANCH}_${short_sha}_$(date +%Y%m%d).tgz"
-tar -czvf "${BUILD_DIR}/${tar_file}" "."
+tar_file="${ELD_BRANCH##*/}_${short_sha}_$(date +%Y%m%d).tgz"
+
+if [[ "${AARCH64_BUILD}" == "true" ]]; then
+    cp -r aarch64/bin aarch64/lib aarch64/libexec aarch64/share aarch64/tools .
```
jonathonpenix: Two thoughts here:

1. By just doing a copy in, I think if the native `bin/`, `lib/`, etc. has _more_ than the aarch64 one, we still have some native libraries/binaries in the final build (I think this is the case in the builds we did earlier--there's some x86 libraries floating around). If we're doing this, we might want to remove the native `bin/`, etc. first (though we need to save the compiler-rt's from the native build). (I'm also assuming we don't have a reason we'd want to mix the two)
2. I'm still not sure if there's any issues since we're building similar projects between the toolchains, but I think it seems safer to make sure we're bringing in everything from the aarch64 one (so `include/` etc.) rather than mixing the two.

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -138,7 +146,42 @@ log "Installing LLVM"
 ninja install
 popd >/dev/null
 
-if [[ "${SKIP_TESTS}" != "true" ]]; then
+if [[ "${AARCH64_BUILD}" == "true" ]]; then
+  log "[Stage 2] Configuring Cross-compiling LLVM for AArch64..."
+  mkdir -p "${BUILD_DIR_AARCH64}" "${INSTALL_DIR_AARCH64}"
+
+  pushd "${BUILD_DIR_AARCH64}" >/dev/null
+  cmake -G Ninja \
```
jonathonpenix, line 154: I'd still lean towards explicitly setting LLVM_HOST_TRIPLE=aarch64-linux-gnu since this was inferred incorrectly IIRC

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -344,26 +382,31 @@ for VARIANT in "aarch64-none-elf" "aarch64-pacret-b-key-bti-none-elf" "armv7-non
...
+    tar_file="${ELD_BRANCH##*/}_${short_sha}_$(date +%Y%m%d).tgz"
+    tar -czvf "${BUILD_DIR}/${tar_file}" "${INSTALL_DIR}"
+fi
 
 if [[ -n "${ARTIFACT_DIR}" ]]; then
-  mkdir -p "${ARTIFACT_DIR}"
-  cp "${BUILD_DIR}/${tar_file}" "${ARTIFACT_DIR}/"
-  log "Artifact copied to ${ARTIFACT_DIR}/${tar_file}"
+    mkdir -p "${ARTIFACT_DIR}"
+    cp "${BUILD_DIR}/${tar_file}" "${ARTIFACT_DIR}/"
```
jonathonpenix: This won't be the right tar file for AArch64, right?

navaneethshan: yes, good catch, thanks.

jonathonpenix (approved):
Before (or when) merging, please be sure to squash down the commits (with description) and include the DCO

Thanks a ton @navaneethshan for this work and cleanup!

jonathonpenix:
Just a misc. note too: I think with this change we're effectively disabling ccache, but I have no issue with that for now (if we want it back in the future, I'd rather we add it via a separate PR--this one is doing enough stuff already)

navaneethshan:
> Just a misc. note too: I think with this change we're effectively disabling ccache, but I have no issue with that for now (if we want it back in the future, I'd rather we add it via a separate PR--this one is doing enough stuff already)

Yes, we need to add it back, will do it as a follow up PR.

PR #35, opened 2025-11-14
Build Linux libc++ configs

jonathonpenix:
Add Linux libc++ configs that relate to the baremetal libc++ configs we are building today.

For better or worse, these roughly match what we were doing in the past with only minor cleanup. There's still lots of weirdness here (relying on libgcc and ld, adding the GNU sysroot to source Linux headers), but we're ignoring that for now with the aim of a) returning to those problems soon and b) keeping this as a relatively small/isolated patch.

### DIFF
--- a/qualcomm-software/embedded/scripts/build.sh
+++ b/qualcomm-software/embedded/scripts/build.sh
@@ -41,6 +41,7 @@ Usage:
 Options:
   --artifact-dir <path>       Directory to copy final tarball
   --skip-tests                Skip LLVM test steps
+  --arm-sysroot <path>        Arm sysroot (default: /usr/arm-linux-gnueabi)
   --aarch64-sysroot <path>    AArch64 sysroot (default: /usr/aarch64-linux-gnu)
   --clean                     Delete and recreate build/install dirs
 
@@ -56,7 +57,8 @@ while [[ $# -gt 0 ]]; do
   case "$1" in
     --artifact-dir) ARTIFACT_DIR="$2"; shift 2 ;;
     --skip-tests) SKIP_TESTS="true"; shift ;;
-    --aarch64-sysroot) AARCH64_SYSROOT="$2"; shift 2 ;;
+    --arm-sysroot) ARM32_SYSROOT_OPT="$2"; shift 2 ;;
+    --aarch64-sysroot) AARCH64_SYSROOT_OPT="$2"; shift 2 ;;
     --clean) CLEAN="true"; shift ;;
     -h|--help) usage; exit 0 ;;
     *) echo "Unknown arg: $1"; usage; exit 1 ;;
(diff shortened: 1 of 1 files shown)


jonathonpenix:
Marking this as a draft--hoping to land https://github.com/qualcomm/cpullvm-toolchain/pull/39 first, then will rebase/fixup these changes

jonathonpenix:
Please take a look--this code will hopefully go away soon, but SriLakshmi asked about merging it.

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -342,8 +347,54 @@ for VARIANT in "aarch64-none-elf" "aarch64-pacret-b-key-bti-none-elf" "armv7-non
...
+
+# Linux libc++ configs
+declare -A LINUX_TRIPLES
+LINUX_TRIPLES["aarch64-linux-gnu"]="aarch64-linux-gnu"
+LINUX_TRIPLES["aarch64-pacret-linux-gnu"]="aarch64-linux-gnu"
+LINUX_TRIPLES["aarch64-pacret-bti-linux-gnu"]="aarch64-linux-gnu"
+LINUX_TRIPLES["armv7-linux-gnueabi"]="armv7-linux-gnueabi"
+declare -A LINUX_CFLAGS
+LINUX_CFLAGS["aarch64-linux-gnu"]="-mcpu=cortex-a53"
+LINUX_CFLAGS["aarch64-pacret-linux-gnu"]="-mcpu=cortex-a53 -march=armv8.3a -mbranch-protection=pac-ret+leaf"
```
navaneethshan, line 361: Aren't "armv8.3-a" the right AArch64 variant name? (is the hypen missing?)

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -342,8 +347,54 @@ for VARIANT in "aarch64-none-elf" "aarch64-pacret-b-key-bti-none-elf" "armv7-non
...
+LINUX_CFLAGS["aarch64-pacret-bti-linux-gnu"]="-mcpu=cortex-a53 -march=armv8.3a -mbranch-protection=pac-ret+leaf+bti"
```
navaneethshan, line 362: Same here.

navaneethshan (approved):
I think both armv8.3-a and armv8.3a are valid variant names.

jonathonpenix:
> I think both armv8.3-a and armv8.3a are valid variant names.

Hm, yeah, I think they should boil down to the same thing (both end up with `"-target-feature" "+v8.3a"`)--looks like we have the same minor inconsistency downstream so I'm not going to worry about it too much for now.

I think the incoming rework is (coincidentally) a bit more consistent in that regard so will leave it be for now and just let the cleanup happen there.

PR #36, opened 2025-11-14
[DNM] Use an explicit sysroot for Arm Linux compiler-rt 

jonathonpenix:
This shouldn't fix anything if someone already had an appropriate cross
toolchain installed (and able to be found by clang). However, this makes
it a bit easier to use sysroots in non-standard locations.


jonathonpenix:
This is stacked on https://github.com/qualcomm/cpullvm-toolchain/pull/35 since the sysroot flag is added there. Once that is merged, this can be rebased, etc.

jonathonpenix:
Just closing this, this code will disappear here soon.

PR #39, opened 2025-11-21
Port workflow cleanup to our main branch

jonathonpenix:
This is basically a cherry-pick of 89bd2c00d0699d6cf12025ebb550c092bc2b8cac except without the nightly build workflow files and without the AArch64 cross build support.

We're heading towards using a single script for everything in mainline so this sort of change needs to happen at some point. Plus, it makes a few patches that are still pending smaller and easier to port to the release branch. So, do this cleanup now.

The nightly build workflow files and AArch64 cross build support are omitted as a) we aren't ready for nightly builds on mainline and b) we don't want the AArch64 cross builds on mainline.


navaneethshan (approved):
LGTM

PR #40, opened 2025-11-26
Make nightly dispatchable

navaneethshan:


.github/workflows/nightly-aarch64.yml
```diff
@@ -1,11 +1,16 @@
...
+    types: [trigger-nightly-aarch64]
+
   schedule:
     - cron: '0 5 * * *' # 9 PM PT (UTC)
 
 jobs:
   base:
     uses: ./.github/workflows/base-build.yml
     with:
-      build-script: "src/qualcomm-software/embedded/scripts/build.sh --artifact-dir /local/mnt/workspace/nightly-toolchain/${{ github.ref_name }} --aarch64-build"
```
jonathonpenix, line 11: Don't we still need the `--aarch64-build` bit?

.github/workflows/nightly-aarch64.yml
```diff
@@ -1,11 +1,16 @@
-name: nightly-aarch64
+name: Nightly-aarch64
```
jonathonpenix: Maybe add Linux suffix here too to be consistent with with x86 and just to keep things clear for when we have Windows builders coming in

.github/workflows/nightly-aarch64.yml
```diff
@@ -1,11 +1,16 @@
...
 
 on:
+  repository_dispatch:
+    types: [trigger-nightly-aarch64]
```
jonathonpenix: Similar thought about adding Linux

.github/workflows/nightly-aarch64.yml
navaneethshan: Ah, yes.

.github/workflows/nightly-aarch64.yml
navaneethshan: Done

.github/workflows/nightly-aarch64.yml
navaneethshan: Done

jonathonpenix (approved):

PR #41, opened 2025-11-28
Enable manual triggering for nightly workflows via workflow_dispatch

navaneethshan:
Added `workflow_dispatch` to nightly-aarch64.yml and nightly-x86-linux.yml
to allow manual runs from GitHub Actions UI and CLI. This makes it possible
to trigger builds on non-default branches without relying on cron or repository_dispatch.


Yethendra (approved):
LGTM

PR #51, opened 2025-12-19
Update ELD to current tip

navaneethshan:
As the branch is synced with upstream llvm-project repo


apazos:
Updating ELD is not the cause of the clang failures. We should allow this change to go in so we can successfully build LLVM + ELD and then we look at the clang failures separately.

Yethendra (approved):

PR #63, opened 2026-01-08
Update compiler-rt patch after recent sync

zhaoshiz:
Resolved conflicts when applying patch


jonathonpenix:
LGTM, please do add the DCO to the commit message before merging though

PR #65, opened 2026-01-11
Refactor build.sh and update ELD

navaneethshan:

PR #34, opened 2025-11-10
Workflow file for Windows build

Yethendra:

PR #66, opened 2026-01-11
Windows workflows and build scripts

navaneethshan:


.github/workflows/nightly-win-x86.yml
```diff
@@ -0,0 +1,42 @@
+name: nightly-win-x86_64
+
+on:
+  pull_request: # removed before merge.
```
navaneethshan: This line needs to be removed.

qualcomm-software/embedded/patches/llvm-project/0001-ARM-compiler-rt-Add-a-few-more-variants-for-builtins.patch
```diff
@@ -1,24 +1,23 @@
-From 0b68523d72053247799419f0f60be7b1835f814e Mon Sep 17 00:00:00 2001
-From: nshanmug <nshanmug@quicinc.com>
-Date: Wed, 17 Sep 2025 13:32:14 -0700
+From 7089c6d2cfb6509b127ee2792dca1f51cc4f3cfe Mon Sep 17 00:00:00 2001
```
navaneethshan: This change will be dropped once https://github.com/qualcomm/cpullvm-toolchain/pull/64 merges.

navaneethshan:
@jonathonpenix, Can you review this PR.

This is everything from https://github.com/qualcomm/cpullvm-toolchain/pull/34 with all the review comments addressed.

qualcomm-software/embedded/scripts/build.ps1
```diff
@@ -0,0 +1,189 @@
...
+# === Config defaults ===
+if (-not $env:CLEAN)          { $env:CLEAN          = "false" }
+if (-not $env:JOBS)           { $env:JOBS           = $env:NUMBER_OF_PROCESSORS }
+if (-not $env:BUILD_MODE)     { $env:BUILD_MODE     = "Release" }
+if (-not $env:ASSERTION_MODE) { $env:ASSERTION_MODE = "OFF" }
+
+# === Constants ===
+$ELD_REPO_URL = "https://github.com/qualcomm/eld.git"
+$ELD_BRANCH   = "main"
+$ELD_COMMIT   = "31c4385321b75c82d03f7244bd5cb4d1d422d1a0"
```
jonathonpenix: If this is the commit we're using here, we should probably be bumping the Linux version as well (separate PR probably)--having two active eld versions is asking for trouble.

qualcomm-software/embedded/scripts/build.ps1
```diff
@@ -0,0 +1,189 @@
...
+  -S "$SRC_DIR\llvm" `
+  -B "$BUILD_DIR\llvm" `
+  -DCMAKE_INSTALL_PREFIX="$INSTALL_DIR" `
+  -DLLVM_TARGETS_TO_BUILD="ARM;AArch64" `
+  -DLLVM_EXTERNAL_PROJECTS="eld" `
+  -DLLVM_EXTERNAL_ELD_SOURCE_DIR="$ELD_DIR" `
+  -DLLVM_DEFAULT_TARGET_TRIPLE="aarch64-unknown-linux-gnu" `
+  -DLIBCLANG_BUILD_STATIC=ON `
+  -DLLVM_POLLY_LINK_INTO_TOOLS=ON `
+  -DCMAKE_C_COMPILER=clang-cl `
```
jonathonpenix, line 154: I guess it looks like we did this for CPULLVM 21.x, but out of curiosity, IIUC in the past downstream we built using Visual Studio --did we have a reason for switching? Or, was there a discussion about this?

qualcomm-software/embedded/scripts/build.ps1
```diff
@@ -0,0 +1,189 @@
...
+  -DCMAKE_CXX_COMPILER=clang-cl `
+  -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL `
+  -DCMAKE_BUILD_TYPE="$env:BUILD_MODE" `
+  -DLLVM_ENABLE_ASSERTIONS="$env:ASSERTION_MODE" `
+  -DLLVM_ENABLE_PROJECTS="llvm;clang;polly;lld;mlir" `
+  -DLLVM_INCLUDE_TESTS=ON `
+  -DLLVM_INSTALL_UTILS=ON `
```
jonathonpenix: Are we adding this to solve a particular issue? Again not a big deal, just curious where this is coming from

qualcomm-software/embedded/scripts/build.ps1
```diff
@@ -0,0 +1,189 @@
...
+  -DLLVM_DEFAULT_TARGET_TRIPLE="aarch64-unknown-linux-gnu" `
+  -DLIBCLANG_BUILD_STATIC=ON `
+  -DLLVM_POLLY_LINK_INTO_TOOLS=ON `
+  -DCMAKE_C_COMPILER=clang-cl `
+  -DCMAKE_CXX_COMPILER=clang-cl `
+  -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL `
+  -DCMAKE_BUILD_TYPE="$env:BUILD_MODE" `
+  -DLLVM_ENABLE_ASSERTIONS="$env:ASSERTION_MODE" `
+  -DLLVM_ENABLE_PROJECTS="llvm;clang;polly;lld;mlir" `
+  -DLLVM_INCLUDE_TESTS=ON `
```
jonathonpenix: Is this added to solve a particular issue? Not a big deal, but should be on by default and it doesn't seem like we explicitly set this in the past.

qualcomm-software/embedded/patches/llvm-project/0001-ARM-compiler-rt-Add-a-few-more-variants-for-builtins.patch
jonathonpenix, line 1: Presumably needs rebase to get rid of this?

qualcomm-software/embedded/scripts/build.ps1
```diff
@@ -0,0 +1,189 @@
...
+# Tool sanity checks
+foreach ($tool in @("git","python","cmake","ninja","clang-cl")) {
+    if (-not (Get-Command $tool -ErrorAction SilentlyContinue)) {
+        Write-Error "*** ERROR: $tool not found on PATH ***"
+        exit 1
+    }
+}
+
+# === Clean ===
+if ($env:CLEAN -eq "true") {
```
jonathonpenix: IIUC, it doesn't look like we set this anywhere, so generally our Windows builders wouldn't explicitly be removing the build/install dir. Is that right?

Since we're using Github runners here (I think), are those clean VMs every time?

Just checking that we aren't going to be "sharing" build/install dirs between builds.

qualcomm-software/embedded/scripts/build.ps1
navaneethshan: yes, its on the way: https://github.com/qualcomm/cpullvm-toolchain/pull/65/changes#diff-2bfed7e6c1014091111c3ecfd59e052344cd595695d4fca3f01b1bcd6ede6736R118

qualcomm-software/embedded/scripts/build.ps1
navaneethshan: ELD is using VS, so I changed to using Visual Studio too, but polly install is failing, I presume its upstream issue, do we need to do polly in Windows? or can you check the logs and suggest?

qualcomm-software/embedded/scripts/build.ps1
navaneethshan: As we using GH hosted Windows server, its clean WS everytime, but you are right, I removed the conditional check, so it we can port it to our own VM as is.

qualcomm-software/embedded/scripts/build.ps1
```diff
@@ -0,0 +1,184 @@
...
+
+# === Prepare workspace ===
+Write-Host "[log] Preparing workspace at: $WORKSPACE"
+New-Item -ItemType Directory -Force -Path $BUILD_DIR,$INSTALL_DIR | Out-Null
+
+# === Clone repos ===
+if (-not (Test-Path "$WORKSPACE\musl-embedded\.git")) {
+    Write-Host "[log] Cloning musl-embedded..."
+    git clone $MUSL_EMBEDDED_REPO_URL "$WORKSPACE\musl-embedded" -b $MUSL_EMBEDDED_BRANCH
+}
```
jonathonpenix, line 96: Actually, it might also be worth skipping the musl-embedded stuff since we don't actually build any of it on windows (not sure if this would mess up patching, etc. though--if this requires any extra work I'd say just leave it as is and we'll deal with it later)

qualcomm-software/embedded/scripts/build.ps1
jonathonpenix: I see, sounds good.

qualcomm-software/embedded/scripts/build.ps1
navaneethshan: If we remove, musl patch apply step will fail, lets deal with it later.

jonathonpenix: Sounds good, thanks for checking

jonathonpenix (approved):
If the Ninja generator + MSVC compiler build is completing successfully, this looks ok to me

(and also if the aarch64 error is unrelated--that needs to be checked first I guess, not sure if you just need to rebase?)

qualcomm-software/embedded/scripts/build.ps1
```diff
@@ -182,7 +183,111 @@ Pop-Location
...
+        "C:\Program Files\7-Zip\7z.exe",
+        "C:\Program Files (x86)\7-Zip\7z.exe"
+    ) | Where-Object { $_ -and (Test-Path $_) } | Select-Object -First 1
+    return $candidates
+}
+
+$usedTool = $null
+
+# Preferred: tar with xz support (Git for Windows ships bsdtar)
+$tarCmd = Get-Command tar -ErrorAction SilentlyContinue
```
jonathonpenix, line 256: FWIW since all this is new, the whole refactor thing coming in (hopefully) soon just delegates packaging to CPack. If this code works as-is, that is fine, but if there are issues it might not be worth the time debugging/trying to make this more/less robust.

Also, not sure what the exact intention is, but for releases my impression is that we should probably default to .zip files for Windows unless we have a pretty good reason not to (or ex: LLVM does an installer + tar archive for development). Unless this is only for storage/internal use, then disregard this bit

qualcomm-software/embedded/scripts/build.ps1
navaneethshan: This has been discussed offline, and we decided to go with clang-cl + Ninja.

PR #70, opened 2026-01-15
Enable RISC-V target in LLVM, eld

jonathonpenix:
`ELD_TARGETS_TO_BUILD` is redundant at this point as we want the same set of targets. Just remove that too while I'm here rather than updating both.

RISC-V Linux runtimes will be enabled in a subsequent PR. Baremetal RISC-V runtimes won't be added for now.


jonathonpenix:
Waiting for https://github.com/qualcomm/cpullvm-toolchain/pull/66/changes to be merged so we can update Windows at the same time

apazos:
LGTM, than you.

PR #75, opened 2026-01-17
[cpullvm] Build RISC-V Linux libraries

jonathonpenix:
Build builtins/compiler-rt/musl/libc++ for RISC-V Linux. Note that this
isn't an exact match of what we've done downstream in the past/what we're
doing for Arm/AArch64. Similar changes for Arm/AArch64 will come in a
follow-up PR.

There's two major ways we're diverging:
  1. The goal here is to build with our own tools and projects that we're
     using rather than leaning on a third party GNU toolchain. This has
     a few important implications for how we're building.
  2. The layout of our final installed libraries is changing. This is
     largely a cosmetic change, but there are some functional differences.

For 1, this means we're changing a few things:
  - We need to source and install the Linux kernel userspace headers (but
    we don't re-distribute these).
  - The order of our builds. As we want our builtins by the time we build
    libc (generally, not for RISC-V here but for Arm/AArch64 or if/when we
    enable shared libraries for RISC-V) but the builtins require C headers
    and the broader compiler-rt build has even more dependencies, we adjust
    how we build projects. Now we do:
      1. Install kernel userspace headers
      2. Install libc headers
      3. Build builtins
      4. Build libc libraries
      5. Build libc++
      6. Build the rest of the compiler-rt libraries
  - Fuzzers are disabled due to what seems like an upstream bug. See inline
    comments for details.

For 2, the new layout looks something like this:
  - libc/libc++: `<install>/<target>/<variant>`
  - compiler-rt: `<resource dir>/lib/<target>/<variant>`

We're doing this partly to align with baremetal (which must change layout),
and partly to align closer towards community (moving away from the
`-<arch>-<arbitrary string>.a` format). This current layout isn't ideal--I
think long term we want something closer to the default Linux/Fuchsia
layout, but there's a few hangups:
  1. That doesn't work particularly well without some form of multilib or
     other source changes to accomodate possibly-conflicting variants with
     the same target triple AFAICT.
  2. There's a few misc. things I'm not sure how to split up by variant
     (libc++ module files, hwasan utilities).
The layout described above lets us more or less sidestep these issues for
now and is relatively close to what we did previously, so it should be
relatively close to what people expect. Eventually, we can/should revisit
this decision.

### DIFF
--- a/qualcomm-software/embedded/scripts/build.sh
+++ b/qualcomm-software/embedded/scripts/build.sh
@@ -444,6 +444,14 @@ for VARIANT in "${!LINUX_TRIPLES[@]}"; do
     log "Linux C++ libs install ..."
 done
 
+# RISC-V Linux libraries
+log "Install RISC-V Linux libraries"
+"${SCRIPT_DIR}"/build_riscv_linux_runtimes.sh \
+                    --base-build-dir "${BUILD_DIR}" \
+                    --base-install-dir "${INSTALL_DIR}" \
+                    --resource-dir "${RESOURCE_DIR}" \
+                    --llvm-src-dir "${SRC_DIR}" \
+                    --download-dir "${WORKSPACE}"
 
 log "Build and installation complete."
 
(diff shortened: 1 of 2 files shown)


jonathonpenix:
https://github.com/qualcomm/cpullvm-toolchain/pull/70 is bundled in here currently, needs rebase eventually.

PR #76, opened 2026-01-20
[cpullvm] Build Arm/AArch64 Linux libraries under the same framework as RISC-V

jonathonpenix:
Arm/AArch64 now follow the same general pattern as RISC-V and are building
against our own tools (llvm tools, musl, etc.)

Note that this again isn't a 1:1 port of our old cpullvm Arm/AArch64
libraries (or what we used to do downstream) even besides the llvm/musl
changes:
* Baseline C flags used to build compiler-rt/musl/libc++ have changed
  slightly--musl used to have slightly different variations, but these
  have been made consistent across the projects.
* The subset of compiler-rt builtins we build for AArch64 Linux is now
  slightly different--we build only the same variants as we build for
  musl/libc++ (rather than sharing some of the same configs with
  baremetal).
* Profile, sanitizer libraries are now building for Arm/AArch64 (they were
  coincidentally disabled in cpullvm previously). There's caveats though,
  see next point.
* We are disabling a few sanitizers/sanitizer-related projects in
  compiler-rt due to either a) upstream issues between LLVM and musl or
  b) our musl version underlying musl-embedded being particularly old. A
  brief survey of the issues (see inline comments for details):
    * sanitizer_common dependent projects have issues for Arm due to
      disagreement about the size of time_t. I don't think this is worth
      fixing, so just disable the projects.
    * rtsan relies on fopencookie, which our musl-embedded is too old to
      support. I don't think this is worth backporting, so disable this.
    * gwp_asan relies on execinfo.h and musl (even upstream musl) does not
      have this header or an equivalent. This is a known issue in upstream
      LLVM. Just disable the project.
* Fix a bug (that also impacts RISC-V!) that meant the extra (non-builtins)
  compiler-rt libraries weren't making it into the final install.

### DIFF
--- a/qualcomm-software/embedded/patches/musl-embedded/0001-Remove-Linux-libraries-from-component_list.sh.patch
+++ b/qualcomm-software/embedded/patches/musl-embedded/0001-Remove-Linux-libraries-from-component_list.sh.patch
@@ -1,0 +1,35 @@
+From 35ba9a719e2dd0015ddae56fc130fe0cbd941be2 Mon Sep 17 00:00:00 2001
+From: Jonathon Penix <jpenix@quicinc.com>
+Date: Sun, 18 Jan 2026 17:48:21 -0800
+Subject: [PATCH] Remove Linux libraries from component_list.sh
+
+We're temporarily porting these over to a consolidated build flow with
+RISC-V prior to a larger refactor. So, don't build these as we'll do it
+elsewhere.
+
+We also don't want to merge this into upstream musl-embedded--we'll be
+removing usage of component_list.sh entirely soon enough. So, just keep
+this as a patch for now.
+
+Signed-off-by: Jonathon Penix <jpenix@quicinc.com>
+---
+ qualcomm-software/config/component_list.sh | 4 ----
+ 1 file changed, 4 deletions(-)
+
+diff --git a/qualcomm-software/config/component_list.sh b/qualcomm-software/config/component_list.sh
+index 2dd84ab1..29e4af4f 100644
+--- a/qualcomm-software/config/component_list.sh
++++ b/qualcomm-software/config/component_list.sh
+@@ -1,9 +1,5 @@
+ # musl_components
+ musl_components=(
+-armv7_linux.sh,armv7-linux-gnueabi
+-aarch64_linux.sh,aarch64-linux-gnu
+-aarch64-pacret_linux.sh,aarch64-pacret-linux-gnu
+-aarch64-pacret-bti_linux.sh,aarch64-pacret-bti-linux-gnu
+ armv7_baremetal.sh,armv7-none-eabi
+ aarch64_baremetal.sh,aarch64-none-elf
+ aarch64-pacret-b-key-bti_baremetal.sh,aarch64-pacret-b-key-bti-none-elf
+-- 
+2.43.0
+
--- a/qualcomm-software/embedded/scripts/build.sh
+++ b/qualcomm-software/embedded/scripts/build.sh
@@ -64,10 +64,8 @@ while [[ $# -gt 0 ]]; do
   case "$1" in
     --artifact-dir) ARTIFACT_DIR="$2"; shift 2 ;;
     --skip-tests) SKIP_TESTS="true"; shift ;;
-    --arm-sysroot) ARM32_SYSROOT_OPT="$2"; shift 2 ;;
     --aarch64-build) AARCH64_BUILD="true"; shift ;;
     --nightly) NIGHTLY="true"; shift ;;
-    --aarch64-sysroot) AARCH64_SYSROOT_OPT="$2"; shift 2 ;;
     --clean) CLEAN="true"; shift ;;
     -h|--help) usage; exit 0 ;;
     *) warn "Unknown arg: $1"; usage; exit 1 ;;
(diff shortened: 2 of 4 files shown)


qualcomm-software/embedded/patches/musl-embedded/0001-Allow-out-of-tree-builds.patch
```diff
@@ -0,0 +1,52 @@
+From 48a39cda1510bd373fb824d90eaf9696ace4347f Mon Sep 17 00:00:00 2001
```
jonathonpenix: patch merged, remove

qualcomm-software/embedded/patches/musl-embedded/0001-Remove-Linux-libraries-from-component_list.sh.patch
```diff
@@ -0,0 +1,36 @@
+From 35ba9a719e2dd0015ddae56fc130fe0cbd941be2 Mon Sep 17 00:00:00 2001
```
jonathonpenix, line 1: This can stay as patch

qualcomm-software/embedded/scripts/build_linux_runtimes.sh
jonathonpenix, line 1: File needs to be renamed

qualcomm-software/embedded/scripts/build_riscv_linux_runtimes.sh
```diff
@@ -89,11 +121,46 @@ popd >/dev/null
 # Variants to build and the basic set of compile flags to use for each. There's
 # surely more elegant ways of doing this, but this doesn't require any extra
 # dependencies and is intended as temporary code.
-VARIANTS=("rv32imac_ilp32" "rv64imac_lp64" "rv64gc_lp64d")
+# VARIANTS=(
```
jonathonpenix: debugging change, ignore

qualcomm-software/embedded/scripts/build_riscv_linux_runtimes.sh
```diff
@@ -304,6 +388,8 @@ for VARIANT in "${VARIANTS[@]}"; do
       -S "${LLVM_BASE_DIR}/runtimes"
   ninja -C "${COMPILER_RT_BUILD_DIR}" install
 
+  exit 1;
```
jonathonpenix: debugging, ignore

jonathonpenix:
FWIW, test-specific flags were also removed as part of this--we weren't running those in cpullvm anyway and there's some work that will be needed to re-enable them. So, I just discarded those flags for now (and it is in our list of TODOs for the release)

qualcomm-software/embedded/scripts/build_linux_runtimes.sh
```diff
@@ -89,11 +121,54 @@ popd >/dev/null
...
+ARCH_MUSL_CFLAGS["aarch64"]="-mstrict-align -fPIC -fno-rounding-math -O3"
+ARCH_MUSL_CFLAGS["arm"]="-mno-unaligned-access -fPIC -fno-rounding-math -O3"
+
+# We also have some new variant-specific configuration going on. Map that out
+# as well. Just list all variants--it'll be cleaned up later.
+declare -A VARIANT_MUSL_CONFIGS
+VARIANT_MUSL_CONFIGS["rv32imac_ilp32"]=""
+VARIANT_MUSL_CONFIGS["rv64imac_lp64"]=""
+VARIANT_MUSL_CONFIGS["rv64gc_lp64d"]=""
+VARIANT_MUSL_CONFIGS["aarch64a"]="--quic-aarch64-optmem"
```
apazos, line 168: Need clarification: if quic_aarch64_optmem is not specified, do we need to still set quic-aarch64-mark-bti?

qualcomm-software/embedded/scripts/build_linux_runtimes.sh
```diff
@@ -263,9 +358,36 @@ for VARIANT in "${VARIANTS[@]}"; do
   ninja -C "${LIBCXX_BUILD_DIR}" install
 
   # Install the rest of compiler-rt now.
+
+  # The goal here is to disable rtsan and gwp_asan:
+  #   * For rtsan, our musl-embedded is too old to support the fopencookie
+  #     extension, which rtsan relies on.
+  #   * For gwp_asan, it requires execinfo.h which is a glibc-specific
+  #     header--no musl version ships this (or an equivalent).
```
apazos, line 366: Double check whether we disabled rtsan in addition to gwp_asan in RISC-V build.

jonathonpenix: Confirming rtsan isn't supported for riscv, so this should be ok--forgot about that earlier, sorry

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -108,10 +98,12 @@ mkdir -p "${BUILD_DIR}" "${INSTALL_DIR}"
 if [[ ! -d "${WORKSPACE}/musl-embedded/.git" ]]; then
   log "Cloning musl-embedded into ${WORKSPACE}/musl-embedded"
   git clone "${MUSL_EMBEDDED_REPO_URL}" "${WORKSPACE}/musl-embedded" -b "${MUSL_EMBEDDED_BRANCH}"
-  MUSL_PINNED_COMMIT="${MUSL_PINNED_COMMIT:-760b7e8ceb9a394e575498ebf8ac79834e9a907c}"
-  pushd "${WORKSPACE}/musl-embedded" >/dev/null
-  git checkout "${MUSL_PINNED_COMMIT}"
-  popd >/dev/null
+  # FIXME: once the musl-embedded changes settle, we need to bump this accordingly
```
jonathonpenix: This needs to be removed and the musl-embedded commit updated, but I don't want to mess with that until support settles.

I'm just removing this until then so that test builds are unblocked.

qualcomm-software/embedded/scripts/build_linux_runtimes.sh
jonathonpenix: This should be fixed after mirroring latest musl-embedded changes (where optmem is enabled for all linux configs).

I guess this is still a question for baremetal configs, but that is a separate discussion I guess.

apazos:
Thanks for the clarification. LGTM

jonathonpenix:
Will merge when builders are done

jonathonpenix:
Our normal precheck is broken/not running, but I'm going to merge this anyway to unblock others--our Linux libraries don't figure into the tests we run right now (llvm/clang/eld/lld/etc. lit tests) so there should be low risk of causing any issues.

PR #78, opened 2026-01-22
Don't use a specific commit for musl-embedded in mainline

jonathonpenix:
Release branches can use a specific commit, but having this in mainline where we want to be using the latest musl-embedded is needlessly fiddly.

This code should all be going away soon with the refactor, so having to re-add this for branches isn't a concern at this point. Note that this is also what the Windows scripts do, and maintaining two of these in mainline isn't worth the effort.


navaneethshan (approved):

apazos:
Sounds reasonable to be on the tip of musl-embedded for the mainline.

PR #85, opened 2026-01-23
Sync 22.x branch with qualcomm-software

navaneethshan:


jonathonpenix:
If you have to make other changes, it might be best to just drop [Don't use a specific commit for musl-embedded in mainline](https://github.com/qualcomm/cpullvm-toolchain/pull/85/changes/f82cd6b811f21c54a5537e0743e66de4cf366306). But not a big deal

Just to double check, if you're merging through the PR, were you planning on doing the rebase option? I think it'd be best to leave these as separate commits

navaneethshan:
> If you have to make other changes, it might be best to just drop [Don't use a specific commit for musl-embedded in mainline](https://github.com/qualcomm/cpullvm-toolchain/pull/85/changes/f82cd6b811f21c54a5537e0743e66de4cf366306). But not a big deal
> 
> Just to double check, if you're merging through the PR, were you planning on doing the rebase option? I think it'd be best to leave these as separate commits

I added the new ELD and MUSL commit, also added back the pinning of musl in the latest commit in this PR.
Also, the plan is to rebase and merge (not squash and merge), so the commits will be merged individually into 22.x.

jonathonpenix:
> I added the new ELD and MUSL commit, also added back the pinning of musl in the latest commit in this PR

Yeah, saw that--I just meant that that commit was originally intended for just main, and _if_ you had to make changes, it might be worth dropping it (and fixing up the shas). But, it's fine as-is if there aren't other changes that need to happen.

> Also, the plan is to rebase and merge (not squash and merge), so the commits will be merged individually into 22.x.

Sounds good, thanks!

jonathonpenix (approved):

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -119,7 +110,7 @@ fi
 if [[ ! -d "${REPO_ROOT}/llvm/tools/eld/.git" ]]; then
   log "Cloning ELD to ${REPO_ROOT}/llvm/tools/eld"
   git clone "${ELD_REPO_URL}" "${SRC_DIR}/llvm/tools/eld" -b "${ELD_BRANCH}"
-  ELD_PINNED_COMMIT="${ELD_PINNED_COMMIT:-96a7dffdf65a68714c8311111d6a6d54a3a150db}"
+  ELD_PINNED_COMMIT="${ELD_PINNED_COMMIT:-c4fd9616c3a95577144b9ae3d9e3f383ccad5156}"
```
jonathonpenix, line 113: Is this the right commit? Should we be on 67fb16670b95badbbae5a01621fe76e8b57a638f instead (since that is eld's release branch IIUC?

Edit: just to rephrase, the c4fd9616c3a95577144b9ae3d9e3f383ccad5156 sha is on both I think so that doesn't make a difference, but should we be catching up to the latest on the release branch? Not a big deal either way probably, just curious

(and the builders have passed, so if we merge and update later that is fine with me)

navaneethshan: Ah, when I sycned this morning, c4fd was the latest on 22.x, I swear :)

May be the commited time you see on the new branch commits are actually main's commit time, and they brought in those commits mins ago, I think.

Anyway we need to sync again soon, so I will go with this now, thanks for checking these Jonathan.

jonathonpenix: Yeah, not sure--sounds good 😄

PR #89, opened 2026-01-25
[cpullvm] Add baseline musl-embedded support, configs

jonathonpenix:
This covers only the basic Arm/AArch64 configs that we added in our initial cpullvm prototype. RISC-V configs will never be added and the various "subvariants" (standalone, uselock, etc.) will be added in a subsequent patch.

Also, build this as part of our precheckin as this is still a fairly core component of our toolchain--long term we can revisit this. I also think the current setup is a bit wasteful in that we build the tools twice, but ccache should make this quick I guess--we can revisit this later.

### DIFF
--- a/.github/workflows/linux-premerge.yml
+++ b/.github/workflows/linux-premerge.yml
@@ -54,5 +54,8 @@ jobs:
       - name: Build
         run: ./qualcomm-software/embedded/scripts/build.sh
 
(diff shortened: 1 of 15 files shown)


apazos (approved):
This might be a subset of MUSL variants we want, but we can add the remaining variants after this patch is merged.

PR #91, opened 2026-01-25
[cpullvm] Add picolibc equivalents of our musl-embedded configs

jonathonpenix:
In theory this way people using the musl configs have a (relatively) direct equivalent that they can try out before we fully deprecate/remove musl.

The pacret configs both have tests disabled for picolibc as there are test failures/faults--these need to be investigated at some point.

Note also that the shared musl/picolibc variant scheme here might fall apart once we add a) musl subvariants (use-locks, etc.) and b) non-TLS/thread/atomic picolibc variants as their multilib custom flags might start to diverge. But, I think that is an issue best sorted out later/separately

### DIFF
--- a/qualcomm-software/embedded/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded/embedded-multilib/json/multilib.json
@@ -8,26 +8,22 @@
         {
             "variant": "aarch64a_soft_nofp",
             "json": "aarch64a_soft_nofp.json",
-            "flags": "--target=aarch64-unknown-none-elf -march=armvX+nofp -march=armvX+nosimd -mabi=aapcs-soft -fno-exceptions",
-            "libraries_supported": "musl-embedded"
+            "flags": "--target=aarch64-unknown-none-elf -march=armvX+nofp -march=armvX+nosimd -mabi=aapcs-soft -fno-exceptions"
         },
         {
             "variant": "aarch64a_soft_nofp_pacret_bti",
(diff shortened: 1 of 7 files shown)


qualcomm-software/embedded/embedded-multilib/json/multilib.json
```diff
@@ -8,26 +8,22 @@
...
         {
             "variant": "aarch64a_soft_nofp_pacret_bti",
             "json": "aarch64a_soft_nofp_pacret_bti.json",
             "flags": "--target=aarch64-unknown-none-elf -march=armv8.3-a -march=armvX+nofp -march=armvX+nosimd -mbranch-protection=pac-ret+leaf+bti -mabi=aapcs-soft -fno-exceptions",
-            "libraries_supported": "musl-embedded"
         },
         {
             "variant": "aarch64a_pacret_bkey_bti",
             "json": "aarch64a_pacret_bkey_bti.json",
             "flags": "--target=aarch64-unknown-none-elf -march=armv8.5-a -mbranch-protection=pac-ret+leaf+b-key+bti -fno-exceptions",
```
apazos: trailing comma, check it does not cause error before merging

apazos (approved):

qualcomm-software/embedded/embedded-multilib/json/multilib.json
jonathonpenix: Removed

PR #99, opened 2026-01-27
Single config that was failing

jonathonpenix:


jonathonpenix:
The issue here is qemu v6.2 has zba/zbb off-by-default

They're also experimental in this version so need to be enabled via x-zba/x-zbb, which is going to cause issues for anyone on newer QEMUs

PR #102, opened 2026-01-28
[cpullvm] Don't ignore fails in the test script

jonathonpenix:
Long term, ignoring fails and dumping the .xml might be better if we process the results and have a convenient way of being alerted to issues. But, we don't. And, for now, having the build outright fail in precheckins, etc. is helpful. So, remove the `--ignore-fails` and xml output flags.

While here, also rename a few things in the precheckin to make it more clear what is running/going on.


jonathonpenix:
Keeping this as a draft as this is expected to fail until we upgrade our qemu version

apazos (approved):
Failing fast now is ok, but in the future we might want a monitoring scheme of failures to avoid situations where no nightly builds are even created.

PR #104, opened 2026-01-28
[cpullvm] Enable x86 target support and build against libstdc++

jonathonpenix:
The x86 change is largely just to enable additional testing and reach
parity with our past releases.

The libstdc++ change has a few motivations:
- Work around test failures when enabling x86 when building against libc++
  that use libunwind.
- This should hopefully be a better experience for most of our expected Linux
  users (on something close to Ubuntu 22+) in that they shouldn't have to
  additionally source libc++ to run the toolchain.

Long term, we'd like to head towards building/packaging a libc++ built from
the same sources, but that's a next step.

x86 support is disabled in eld for now as there are test failures when
building/testing on native AArch64 machines. This can be re-enabled when
https://github.com/qualcomm/eld/issues/764 is resolved.

### DIFF
--- a/qualcomm-software/embedded/CMakeLists.txt
+++ b/qualcomm-software/embedded/CMakeLists.txt
@@ -178,11 +178,14 @@ set(LLVM_TOOLCHAIN_DISTRIBUTION_COMPONENTS
 installed by the install-llvm-toolchain target"
 )
 set(LLVM_ENABLE_PROJECTS clang;lld;polly CACHE STRING "")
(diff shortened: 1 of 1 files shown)


jonathonpenix:
(we made the [same change in our branch already](https://github.com/qualcomm/cpullvm-toolchain/commit/5e23c1861580af870fd2f21c85b288fd212e7bbe), just bringing this into qualcomm-software/the refactor now)

(see also: https://github.com/qualcomm/cpullvm-toolchain/pull/106 disabling eld in the branch)

apazos (approved):

PR #105, opened 2026-01-29
Verify AArch64 native build

navaneethshan:


.github/workflows/base-build.yml
```diff
@@ -12,6 +12,9 @@ on:
       arch:
         required: false
         type: string
+      runner:
+        required: false
```
navaneethshan: Make this true.

qualcomm-software/embedded/scripts/build.sh
```diff
@@ -126,13 +122,23 @@ python3 "${SRC_DIR}/qualcomm-software/embedded/tools/patchctl.py" apply -f "${SR
...
+fi
+
 cmake -G Ninja -DCMAKE_INSTALL_PREFIX="${INSTALL_DIR}" \
   -DLLVM_TARGETS_TO_BUILD="ARM;AArch64;RISCV;X86" \
-  -DELD_TARGETS_TO_BUILD="AArch64;ARM;RISCV" \
   -DLLVM_EXTERNAL_PROJECTS="eld" \
   -DLLVM_EXTERNAL_ELD_SOURCE_DIR="llvm/tools/eld" \
-  -DLLVM_DEFAULT_TARGET_TRIPLE="aarch64-unknown-linux-gnueabi" \
-  -DLLVM_TARGET_ARCH="arm-linux-gnueabi" \
+  -DLLVM_DEFAULT_TARGET_TRIPLE="${DEFAULT_TRIPLE}" \
```
jonathonpenix: Two thoughts:
* Do we want the different default targets on different hosts? We've set it to arm/aarch64 in the past (which, I guess is fair enough for what is intended to be a cross compiler).
* If we want the default target to match the host, we can just omit this and LLVM's cmake will pick 'host'/the appropriate triple (or we could directly set host I guess)

navaneethshan: Let me try the second option.

navaneethshan: Well, I brought back what we had.

jonathonpenix: Sounds good

Just out of curiosity, did the the 'host' thing not work? Just asking as I see the "Add LLVM_DEFAULT_TARGET_TRIPLE back" commit--curious if something went wrong or you just decided to keep the old code

jonathonpenix (approved):

qualcomm-software/embedded/scripts/build.sh
navaneethshan: No reason, I just decided to have the old flag as is.

PR #106, opened 2026-01-29
[cpullvm] Disable x86 support in eld

jonathonpenix:
Currently there are test failures when running eld's x86 tests on native AArch64 Linux machines. For now, just disable eld x86 support as x86 is largely intended as a debugging/convenience tool. This can be re-enabled when https://github.com/qualcomm/eld/issues/764 is resolved.


apazos (approved):

navaneethshan (approved):
LGTM

PR #109, opened 2026-01-30
[cpullvm] Expand usages of 'musl' to 'musl-embedded'. NFCI

jonathonpenix:
I thought there were some sharp edges around '-' in variable names [1][2] in cmake so originally I tried to avoid this sort of thing when adding musl-embedded support. But, based on [3] I think this is actually fine for what we do (and experimentally things seem to be working as expected). We're getting ready to add upstream musl as a dependency for our linux configs, so the shorthand 'musl' to refer to musl-embedded needs to change. Just use 'musl-embedded' in full to make things as clear as possible.

[1] https://cmake.org/cmake/help/book/mastering-cmake/chapter/Writing%20CMakeLists%20Files.html#variables
[2] https://gitlab.kitware.com/cmake/cmake/-/issues/23075
[3] https://cmake.org/cmake/help/v3.20/manual/cmake-language.7.html#variables

### DIFF
--- a/qualcomm-software/embedded/cmake/generate_version_txt.cmake
+++ b/qualcomm-software/embedded/cmake/generate_version_txt.cmake
@@ -25,11 +25,7 @@ execute_process(
 )
 
 # Supported libcs are all in a separate repo
-if(LLVM_TOOLCHAIN_C_LIBRARY STREQUAL musl-embedded)
-    set(base_library musl)
-else()
-    set(base_library ${LLVM_TOOLCHAIN_C_LIBRARY})
-endif()
+set(base_library ${LLVM_TOOLCHAIN_C_LIBRARY})
 
 execute_process(
     COMMAND git -C ${${base_library}_SOURCE_DIR} rev-parse HEAD
--- a/qualcomm-software/embedded/embedded-multilib/json/variants/aarch64a.json
+++ b/qualcomm-software/embedded/embedded-multilib/json/variants/aarch64a.json
@@ -26,7 +26,7 @@
             "ENABLE_LIBC_TESTS": "OFF",
             "ENABLE_COMPILER_RT_TESTS": "OFF",
             "ENABLE_LIBCXX_TESTS": "OFF",
-            "EXTRA_MUSL_CFLAGS": "-mstrict-align -D__QUIC_ENABLE_FLT_FOR_PRINT"
+            "EXTRA_MUSL-EMBEDDED_CFLAGS": "-mstrict-align -D__QUIC_ENABLE_FLT_FOR_PRINT"
         }
     }
 }
--- a/qualcomm-software/embedded/embedded-runtimes/CMakeLists.txt
+++ b/qualcomm-software/embedded/embedded-runtimes/CMakeLists.txt
@@ -86,12 +86,12 @@ set(ENABLE_COMPILER_RT_TESTS ${ENABLE_COMPILER_RT_TESTS_def} CACHE BOOL "Enable 
 set(ENABLE_LIBCXX_TESTS ${ENABLE_LIBCXX_TESTS_def} CACHE BOOL "Enable libcxx tests.")
 set(LLVM_BINARY_DIR "" CACHE PATH "Path to LLVM toolchain root to build libraries with")
 
-# Set up a few arbitrary pass-through flags for MUSL. This probably isn't the best idea,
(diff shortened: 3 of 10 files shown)


qualcomm-software/embedded/embedded-multilib/json/variants/aarch64a.json
```diff
@@ -26,7 +26,7 @@
             "ENABLE_LIBC_TESTS": "OFF",
             "ENABLE_COMPILER_RT_TESTS": "OFF",
             "ENABLE_LIBCXX_TESTS": "OFF",
-            "EXTRA_MUSL_CFLAGS": "-mstrict-align -D__QUIC_ENABLE_FLT_FOR_PRINT"
+            "EXTRA_MUSL-EMBEDDED_CFLAGS": "-mstrict-align -D__QUIC_ENABLE_FLT_FOR_PRINT"
```
apazos, line 29: we had flags like QUIC_BAREMETAL in musl-embedded, I do not see them in the list of flags.

qualcomm-software/embedded/embedded-runtimes/CMakeLists.txt
```diff
@@ -567,10 +567,10 @@ if(C_LIBRARY STREQUAL musl-embedded)
         --disable-wrapper
         --quic-arm-baremetal
```
apazos, line 568: Probably it is this one --quic-arm-baremetal that causes the QUIC_BAREMETAL macro to be enabled.

apazos (approved):

apazos:
@zhaoshiz, could you kindly review this patch.

qualcomm-software/embedded/cmake/generate_version_txt.cmake
```diff
@@ -25,11 +25,7 @@ execute_process(
 )
 
 # Supported libcs are all in a separate repo
-if(LLVM_TOOLCHAIN_C_LIBRARY STREQUAL musl-embedded)
-    set(base_library musl)
```
zhaoshiz, line 29: can you elaborate this part? just to make sure it's NFC

qualcomm-software/embedded/embedded-runtimes/CMakeLists.txt
zhaoshiz: yes, `--quic-arm-baremetal` is passed to `configure` where `-DQUIC_BAREMEAL` is added to CFLAGS

qualcomm-software/embedded/embedded-runtimes/CMakeLists.txt
zhaoshiz, line 588: @apazos here is the command to call `configure` with `musl_embedded_flags` defined above

zhaoshiz (commented):
Overall LGTM, just need to make sure changes to qualcomm-software/embedded/cmake/generate_version_txt.cmake is NFC.

qualcomm-software/embedded/cmake/generate_version_txt.cmake
jonathonpenix: There used to be a mismatch between what `LLVM_TOOLCHAIN_C_LIBRARY` contained (musl-embedded) and the various  `<libname>_SOURCE_DIR`/`_URL`/etc. variables (libname == musl). This just did the mapping from musl-embedded -> musl so we can handle it consistently with other libraries (picolibc) below where they're the same.

At this point, all the libcs we support will have LLVM_TOOLCHAIN_C_LIBRARY equal to libname in <libname>_SOURCE_DIR so we can just remove this for now. (Arm's toolchain still does this for ex: newlib nano: https://github.com/arm/arm-toolchain/blob/8881f72d070e4556e3e9dd491e35f6dc2b946177/arm-software/embedded/cmake/generate_version_txt.cmake#L21-L25, but we don't support newlib.

qualcomm-software/embedded/embedded-multilib/json/variants/aarch64a.json
jonathonpenix: Your comment below is correct I think (it propagates out from the `--quic...` flag)

qualcomm-software/embedded/cmake/generate_version_txt.cmake
zhaoshiz: Thanks for the clarification @jonathonpenix

zhaoshiz (approved):

PR #110, opened 2026-01-30
[cpullvm] Move qualcomm-software/embedded/* to qualcomm-software/. NFCI

jonathonpenix:
The embedded folder isn't a helpful distinction--as we start to add Linux libraries, we want both Linux and embedded to be part of the same project. As such, I think it makes more sense to have them at the same "level" in our folder.

To prepare for this, move everything out of qualcomm-software/embedded into qualcomm-software.


apazos (approved):

jonathonpenix:
Whoops, these build failures are correct--I totally forgot about the builder. Will fix in a minute

PR #115, opened 2026-02-01
[cpullvm] Add AArch64 Linux build to our premerge builds

jonathonpenix:
There's dependencies that need to be sorted out in the AArch64 builder (meson, etc.). So for now, just disable building the runtimes on AArch64. build_no_runtimes.sh should be deleted once this is resolved.

Also add an EXTRA_CMAKE_ARGS passthrough in our scripts--this is helpful even when working locally to modify things as appropriate.


apazos (approved):
It sounds fine to skip building the runtimes in any other builder other than linux x86.

PR #127, opened 2026-02-04
[cpullvm][Github] Remove explicit clean steps from workflows

jonathonpenix:
Our Linux workflows have generally had a "clean" step as our self hosted builders would otherwise be doing incremental builds between workflows, which isn't good. Our old Windows build.ps1 scripts also did this to some degree pre-emptively (currently we use Github runners which are "fresh" run-to-run).

After the refactor, our builds changed slightly--rather than defaulting to building *out* of the repo (so build/ install/ cpullvm-toolchain/ are all separate), our build scripts now prefer builds *in* the repo (cpullvm-toolchain/build* is the path used). As a result, when 'clean' is true when running actions/checkout (it is true by default, and we explicitly set it to true in some workflows), the various build folders will be cleaned up as part of the checkout step.

I think relying on this behavior is nice in that it lets us handle Linux/Windows and self-hosted/Github runners consistently in this regard with less code.

### DIFF
--- a/.github/workflows/linux-premerge.yml
+++ b/.github/workflows/linux-premerge.yml
@@ -53,19 +53,13 @@ jobs:
             runner: self-hosted
 
     steps:
-      - name: Cleanup workspace
-        shell: bash
-        run: |
-          set -euo pipefail
-          echo "Cleaning workspace: ${GITHUB_WORKSPACE}"
-          shopt -s dotglob nullglob
-          rm -rf "${GITHUB_WORKSPACE:?}/"*
-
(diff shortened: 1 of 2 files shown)


jonathonpenix:
For reference, see https://github.com/actions/checkout/blob/0c366fd6a839edf440554fa01a7085ccba70ac98/README.md?plain=1#L111-L113 on the behavior of `clean` ("Whether to execute `git clean -ffdx && git reset --hard HEAD`")

navaneethshan (approved):

PR #130, opened 2026-02-04
[cpullvm] Re-add step to configure git, re-enable tests for Windows nightlies

jonathonpenix:
I missed this change when porting our old workflows. Without this, we see lit
test failures as files are comparing differently due to different line endings.

With this change in place, Windows tests are passing so re-enable them in the
nightlies.

Note that this change is done for both Linux and Windows, just for simplicity.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -15,7 +15,7 @@
 name: Nightly Build and Test
 
 on:
-  workflow_dispatch: 
+  workflow_dispatch:
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


jonathonpenix:
This seems to be the minimal change required, see: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/21737668444/job/62706146626?pr=130

apazos (approved):
Thanks Jonathon for tracking and resolving the windows issues!

navaneethshan (approved):
LGTM, thanks for posting the test job link.

PR #133, opened 2026-02-05
[cpullvm] Add lldb to the list of projects to build/distribute

jonathonpenix:
We have a few issues with integrating lldb at the moment:
* lldb seems to expect to be built as a top-level project while we add it (with
  the rest of llvm) via `add_subdirectory`. This is causing lots of issues with
  the python support and tests.
* lldb tests don't seem particularly happy about a default target triple that
  doesn't necessarily match the host--binaries are built using the default
  target triple and executed, which doesn't end well.
We have workarounds for the second issue, but not so much the first. So, just
disable testing and python integration for now while we sort these out.

The other thing to note here is that most of our builders are missing most of
lldb's optional dependencies (libedit, etc.). These are intentionally being
ignored while we a) figure out which ones we care about and b) add them to our
builders.

### DIFF
--- a/qualcomm-software/CMakeLists.txt
+++ b/qualcomm-software/CMakeLists.txt
@@ -167,8 +167,15 @@ set(LLVM_DISTRIBUTION_COMPONENTS
     clang-scan-deps
     clang-tidy
     clangd
-    lld
     ld.eld
+    liblldb
+    lld
+    lldb
+    lldb-argdumper
+    lldb-dap
+    lldb-instr
+    lldb-mcp
+    lldb-server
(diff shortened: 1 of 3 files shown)


apazos (approved):

apazos (approved):

jonathonpenix:
Windows failures are unrelated, they show up in eld's nightly as well: https://github.com/qualcomm/eld/actions/runs/22700071294

Will remove nightly line and merge.

navaneethshan:
>/cherry-pick deb2757

Error: Command failed due to missing milestone.

PR #134, opened 2026-02-05
[cpullvm] Use eld throughout the Linux runtime build

jonathonpenix:
We're more consistent this way, once we re-enable tests it is good coverage for eld, and xqci configs need a linker that knows about their relocations.


jonathonpenix:
Hrm, seems we can't do this right now, see https://github.com/qualcomm/eld/issues/786

jonathonpenix:
This is building now, though there's still some work on this front on the sysroot thing in eld I think: https://github.com/qualcomm/eld/pull/809

Flipping everything over to eld might be an exercise best left for mainline post-release? (mostly in case there are any issues only seen at runtime, etc.)

jonathonpenix:
No plans to move forward with this, closing.

PR #136, opened 2026-02-05
[cpullvm] Add aarch64 pacret + b-key + bti Linux variant

jonathonpenix:
In the past, this was compiler-rt only but for now let's build the full compiler-rt/libc/libc++ set.

### DIFF
--- a/qualcomm-software/scripts/build_linux_runtimes.sh
+++ b/qualcomm-software/scripts/build_linux_runtimes.sh
@@ -118,6 +118,7 @@ VARIANTS=(
   "aarch64a"
   "aarch64a_pacret"
   "aarch64a_pacret_bti"
+  "aarch64a_pacret_bkey_bti"
   "armv7_softfp_neon"
 )
 
@@ -137,6 +138,7 @@ VARIANT_BUILD_FLAGS["rv64gc_lp64d"]="--target=riscv64-unknown-linux-gnu -march=r
 VARIANT_BUILD_FLAGS["aarch64a"]="--target=aarch64-unknown-linux-gnu -mcpu=cortex-a53"
 VARIANT_BUILD_FLAGS["aarch64a_pacret"]="--target=aarch64-unknown-linux-gnu -mcpu=cortex-a53 -march=armv8.3a -mbranch-protection=pac-ret+leaf"
 VARIANT_BUILD_FLAGS["aarch64a_pacret_bti"]="--target=aarch64-unknown-linux-gnu -mcpu=cortex-a53 -march=armv8.3a -mbranch-protection=pac-ret+leaf+bti"
+VARIANT_BUILD_FLAGS["aarch64a_pacret_bkey_bti"]="--target=aarch64-unknown-linux-gnu -mcpu=cortex-a53 -march=armv8.5a -mbranch-protection=pac-ret+leaf+b-key+bti"
 # The full normalized form here seems to be "thumbv7-unknown-linux-gnueabi" but
 # that doesn't seem to be what clang searches--use "arm-unknown-linux-gnueabi"
 # instead.
@@ -161,6 +163,8 @@ VARIANT_MUSL_CONFIGS["aarch64a"]="--quic-aarch64-optmem"
 VARIANT_MUSL_CONFIGS["aarch64a_pacret"]="--quic-aarch64-optmem"
 VARIANT_MUSL_CONFIGS["aarch64a_pacret_bti"]="--quic-aarch64-optmem \
                                              --quic-aarch64-mark-bti"
+VARIANT_MUSL_CONFIGS["aarch64a_pacret_bkey_bti"]="--quic-aarch64-optmem \
+                                                  --quic-aarch64-mark-bti"
 VARIANT_MUSL_CONFIGS["armv7_softfp_neon"]=""
 
 for VARIANT in "${VARIANTS[@]}"; do


qualcomm-software/scripts/build_linux_runtimes.sh
```diff
@@ -161,6 +163,8 @@ VARIANT_MUSL_CONFIGS["aarch64a"]="--quic-aarch64-optmem"
 VARIANT_MUSL_CONFIGS["aarch64a_pacret"]="--quic-aarch64-optmem"
 VARIANT_MUSL_CONFIGS["aarch64a_pacret_bti"]="--quic-aarch64-optmem \
                                              --quic-aarch64-mark-bti"
+VARIANT_MUSL_CONFIGS["aarch64a_pacret_bkey_bti"]="--quic-aarch64-optmem \
+                                                  --quic-aarch64-mark-bti"
```
jonathonpenix, line 167: @pzhengqc I'm kind of guessing here, but I assume we'd want the optmem + mark bti here? Can you confirm?

zhaoshiz: yes, IIRC, we decided to enable optimized memcpy/memset for aarch64 and bti variants requires mark-bti

qualcomm-software/scripts/build_linux_runtimes.sh
```diff
@@ -137,6 +138,7 @@ VARIANT_BUILD_FLAGS["rv64gc_lp64d"]="--target=riscv64-unknown-linux-gnu -march=r
 VARIANT_BUILD_FLAGS["aarch64a"]="--target=aarch64-unknown-linux-gnu -mcpu=cortex-a53"
 VARIANT_BUILD_FLAGS["aarch64a_pacret"]="--target=aarch64-unknown-linux-gnu -mcpu=cortex-a53 -march=armv8.3a -mbranch-protection=pac-ret+leaf"
 VARIANT_BUILD_FLAGS["aarch64a_pacret_bti"]="--target=aarch64-unknown-linux-gnu -mcpu=cortex-a53 -march=armv8.3a -mbranch-protection=pac-ret+leaf+bti"
```
zhaoshiz, line 140: BTI is added in armv8.5a, we build compiler-rt using `8.5a` for all pacret[-b-key] variants regardless of bti in our downstream toolchain. But for musl we use `8.3a` for pacret without bti and `8.5a` for bti variants.

I suggest we either use `8.5a` for all pacret[-b-key][-bti], or use `8.3a` for pacret, and `8.5a` for pacret[-b-key]-bti.

jonathonpenix: > I suggest we either use 8.5a for all pacret[-b-key][-bti], or use 8.3a for pacret, and 8.5a for pacret[-b-key]-bti.

I'm fine with this change in general--I'm not at all familiar with these, but it seems like that makes sense to me?

But, just to clarify:

> BTI is added in armv8.5a, we build compiler-rt using 8.5a for all pacret[-b-key] variants regardless of bti in our downstream toolchain. But for musl we use 8.3a for pacret without bti and 8.5a for bti variants.

Agreed on the compiler-rt front, but for musl, my understanding is we _did_ use 8.3a for bti--see https://github.com/qualcomm/musl-embedded/blob/main/qualcomm-software/config/linux/arm/aarch64-pacret-bti_linux.sh or https://github.com/qualcomm/musl-embedded/blob/main/qualcomm-software/config/linux/arm/aarch64-pacret-bti_baremetal.sh (and, this is true downstream as well). This mismatch is how I ended up with the 8.3 here--do you know if there is a reason we used to use 8.3 or is that likely just an accident?

jonathonpenix: I guess the other thing that confuses me is Arm ships 8.1m-main with bti enabled, ex: https://github.com/arm/arm-toolchain/blob/26e1a3d5f3fc0e86f7e1c96821ea289f32d19f81/arm-software/embedded/arm-multilib/json/multilib.json#L539-L541

So I'm generally a bit confused on how bti fits into everything?

qualcomm-software/scripts/build_linux_runtimes.sh
pzhengqc: Yes, @jonathonpenix, the flags look correct here. We need optmem + mark-bti here.

qualcomm-software/scripts/build_linux_runtimes.sh
zhaoshiz: >do you know if there is a reason we used to use 8.3 or is that likely just an accident?

yes, looks like we have some inconsistency between compiler-rt and musl downstream

> I guess the other thing that confuses me is Arm ships 8.1m-main with bti enabled, ex: https://github.com/arm/arm-toolchain/blob/26e1a3d5f3fc0e86f7e1c96821ea289f32d19f81/arm-software/embedded/arm-multilib/json/multilib.json#L539-L541

This seems for ARM32 (--target=thumbv8.1m...), pacret-bti is added to 8.1m-main.

Since we only enable pacret[-bti] for AArch64, let's use 8.3a for pacret without bti, and 8.5a for pacret[-b-key]-bti

jonathonpenix: Sounds good! Please see https://github.com/qualcomm/cpullvm-toolchain/pull/137

qualcomm-software/scripts/build_linux_runtimes.sh
jonathonpenix: Thanks Zhaoshi and Pengxuan!

pzhengqc (approved):

apazos (approved):

apazos:
Bumping up to 8.5 is handled in separate patch

PR #137, opened 2026-02-05
[cpullvm] Change AArch64 BTI variants to use v8.5a

jonathonpenix:
In the past we had a mix where some components (compiler-rt) would use v8.5a and others would use v8.3a so for cpullvm I had just aligned down. It seems this mismatch likely wasn't intentional, so let's align up to v8.5a.


zhaoshiz (approved):
Thanks for fixing this!

pzhengqc (approved):

apazos (approved):

jonathonpenix:
Just rebasing--will merge if/when builders are clean

PR #139, opened 2026-02-06
Add riscv32imac_ilp32 baremetal runtime variant

pzhengqc:
This patch adds the riscv32imac_ilp32 PIC variant.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -26,6 +26,12 @@
             "flags": "--target=thumbv7-unknown-none-eabi -mfpu=neon -fno-exceptions"
         },
         {
+            "variant": "riscv32imac_ilp32",
+            "json": "riscv32imac_ilp32.json",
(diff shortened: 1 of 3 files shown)


jonathonpenix:
Please also do add tests into 'qualcomm-software/test/multilib' to make sure we are selecting the right multilib variants (and, I think the build failure is just that one of the tests checking that we _don't_ provide a PIC-compatible variant needs to be updated)

You can run only the multilib tests locally with `ninja check-llvm-toolchain-lit`

pzhengqc:
> Please also do add tests into 'qualcomm-software/test/multilib' to make sure we are selecting the right multilib variants (and, I think the build failure is just that one of the tests checking that we _don't_ provide a PIC-compatible variant needs to be updated)
> 
> You can run only the multilib tests locally with `ninja check-llvm-toolchain-lit`

Thanks, @jonathonpenix! The build failure is indeed due to the test failing. I have updated the test accordingly.

jonathonpenix (approved):
Looks good, thank you!

PR #141, opened 2026-02-07
Add riscv64gc_lp64d_nopic baremetal runtime variant

pzhengqc:
This patch adds the riscv64gc_lp64d_nopic variant. We xfail two picolibc tests (math_errhandling.test and test-fma.test) for now. These are known issues and being fixed upstream.

PR #142, opened 2026-02-07
Fix riscv64gc_lp64d_nopic variant build error

pzhengqc:
While adding the riscv64gc_lp64d_nopic variant, the following build error was observed.
```
In file included from
/local/mnt/workspace/actions-runner-2/_work/cpullvm-toolchain/cpullvm-toolchain/libunwind/src/Unwind-wasm.c:15: /local/mnt/workspace/actions-runner-2/_work/cpullvm-toolchain/cpullvm-toolchain/libunwind/src/config.h:16:10: fatal error: 'assert.h' file not found
   16 | #include <assert.h>
      |          ^~~~~~~~~~
1 error generated.
```
Adding `-isystem` fixes the error.

### DIFF
--- a/qualcomm-software/embedded-runtimes/CMakeLists.txt
+++ b/qualcomm-software/embedded-runtimes/CMakeLists.txt
@@ -203,7 +203,10 @@ if(ENABLE_LIBC_TESTS OR ENABLE_COMPILER_RT_TESTS OR ENABLE_LIBCXX_TESTS)
     list(JOIN lit_test_executor " " lit_test_executor)
 endif()
 
-set(compile_arch_flags "--target=${target_triple} ${COMPILE_FLAGS} --sysroot ${TEMP_LIB_DIR}")
+# Adding "-isystem ${TEMP_LIB_DIR}/include" to work around a sysroot issue for
+# RISC-V baremetal. We may not need this workaround in the future if the issue
+# is fixed upstream.
+set(compile_arch_flags "--target=${target_triple} ${COMPILE_FLAGS} --sysroot ${TEMP_LIB_DIR} -isystem ${TEMP_LIB_DIR}/include")
 # Compiling the libraries benefits from some extra optimization
 # flags, and requires a sysroot.
 set(lib_compile_flags "${compile_arch_flags} -ffunction-sections -fdata-sections")


pzhengqc:
Here's a link to the failed build:
https://github.com/qualcomm/cpullvm-toolchain/actions/runs/21771087447/job/62818616910?pr=141#step:4:62391

jonathonpenix (changes requested):
Removing sysroot and replacing it with just isystem isn't right--sysroot is doing more than that (helping to find libraries, etc.)

More generally, this doesn't seem like the right fix--this would imply something is broken with sysroot I think? (or there's some weird include order shenanigans, but I'd be surprised if this is just appearing now)

jonathonpenix:
I poked at this a bit and I think I see what is happening--I think we're running into issues with the RISC-V hardcoded multilib (which, would make sense that -isystem worked around this).

@pzhengqc can you please take a look at/try building with https://github.com/qualcomm/cpullvm-toolchain/pull/143 and let me know if it works? `check-clang` will probably fail as I didn't touch the tests at all, but the build should hopefully succeed.

pzhengqc:
> Removing sysroot and replacing it with just isystem isn't right--sysroot is doing more than that (helping to find libraries, etc.)
> 
> More generally, this doesn't seem like the right fix--this would imply something is broken with sysroot I think? (or there's some weird include order shenanigans, but I'd be surprised if this is just appearing now)

I was trying to see if replacing `--sysroot` with `-isystem` is enough since we are mostly just building static libraries for baremetal. Looks like we can not remove `--sysroot` as picolibc seems to build some test executables too which requires linking with libraries. I guess we can just add `-isysroot` and keep `--sysroot`.

pzhengqc:
> I poked at this a bit and I think I see what is happening--I think we're running into issues with the RISC-V hardcoded multilib (which, would make sense that -isystem worked around this).
> 
> @pzhengqc can you please take a look at/try building with #143 and let me know if it works? `check-clang` will probably fail as I didn't touch the tests at all, but the build should hopefully succeed.

Will try #143.

jonathonpenix:
> > Removing sysroot and replacing it with just isystem isn't right--sysroot is doing more than that (helping to find libraries, etc.)
> > More generally, this doesn't seem like the right fix--this would imply something is broken with sysroot I think? (or there's some weird include order shenanigans, but I'd be surprised if this is just appearing now)
> 
> I was trying to see if replacing `--sysroot` with `-isystem` is enough since we are mostly just building static libraries for baremetal. Looks like we can not remove `--sysroot` as picolibc seems to build some test executables too which requires linking with libraries. I guess we can just add `-isysroot` and keep `--sysroot`.

I don't think we want both--the isysroot fixes the issue of finding picolibc headers, but sysroot will still be broken when it comes to finding the libraries for these configs (I think it'll still look into `sysroot/rv64gc/lp64d/lib` or whatever rather than `sysroot/lib`) and I think we'd need to handle C++ headers in the same way if/when we ex: try to enable compiler-rt tests on baremetal. At that point I think we're better off resolving the underlying issue (which, #143 might not be the right way to do that, FWIW) so we can treat Arm/AArch64 and RISC-V consistently rather than pass through the various `isystem`/`-L` flags to force this to work correctly.

qualcomm-software/embedded-runtimes/CMakeLists.txt
```diff
@@ -203,7 +203,7 @@ if(ENABLE_LIBC_TESTS OR ENABLE_COMPILER_RT_TESTS OR ENABLE_LIBCXX_TESTS)
     list(JOIN lit_test_executor " " lit_test_executor)
 endif()
 
-set(compile_arch_flags "--target=${target_triple} ${COMPILE_FLAGS} --sysroot ${TEMP_LIB_DIR}")
+set(compile_arch_flags "--target=${target_triple} ${COMPILE_FLAGS} --sysroot ${TEMP_LIB_DIR} -isystem ${TEMP_LIB_DIR}/include")
```
jonathonpenix, line 209: We discussed this today and decided to stay with `-isystem`/etc. as a temporary solution pending looking at using a placeholder mutlilib.yaml and/or some solution in upstream.

Two things:
* Can you please add a comment here on why the -isystem (and/or other flags) is necessary?
* Can you please post a link of a build with this and the lp64d build? I'm still assuming we're going to need more than just isystem as ex: picolibc tests still need to find the builtins, etc. and this issue isn't specific to include directories.)

pzhengqc: You mean trigger a build with this PR and #141? Can you please explain how to trigger this build? I did a local build (running qualcomm-software/scripts/build.sh) with both patches and didn't run into any other build errors.

jonathonpenix: Hm, I guess the easiest way is just to push a draft PR with both commits--I'm not sure there's a better way right now 😕 

Also, FWIW, another thing to keep an eye on is picolibc uses the builtins in it's link tests, ex: https://github.com/picolibc/picolibc/blob/11306106de548ac419c4c776f07327a32b05d855/meson.build#L212C1-L226 (and we set that via meson-cross-build.txt.in). We might not see hard errors, but worth keeping an eye on whether our configure "results" change (I've broken things in MUSL in this same way)

pzhengqc: The combined verification #154 is clean now after disabling two picolibc tests which are known issues according to @jonathonpenix. I will update #141 to disable the two picolibc tests. I think this patch is ready to be merged now.

jonathonpenix: Thanks for testing this @pzhengqc!

jonathonpenix (approved):

apazos (approved):

PR #143, opened 2026-02-07
[cpullvm][RISCV] Remove hardcoded RISC-V multilib

jonathonpenix:
When we build our embedded runtimes, we start with an empty directory as our sysroot and pass `--sysroot <dir>` to each step as we build it out. Notably, we don't have a multilib.yaml at this stage or rely on multilib at all during this process. Also note that the sysroot contains only a 'include' and 'lib' dir that we build out and expect the usual clang driver logic to handle. This scheme works nicely for Arm and AArch64.

For RISC-V, we can run into issues with the hardcoded multilibs. For certain configs (-march=rv64gc -mabi=lp64d, for example) we match the hardcoded multilib (and again we don't have a multilib.yaml to disable this). As such, when we pass in `--sysroot <dir>`, instead of finding files in `<dir>/include` and `<dir>/lib` as normal, we look in paths like
`<dir>/rv64imafdc/lp64d/include` which don't exist, causing build failures (failed to find #include'd headers, etc.)

We could maybe install a dummy multilib.yaml to disable this behavior, but that seems like more hassle than it is worth. And, these hardcoded multilib paths are not valid in our toolchain and we have no intention of adopting that scheme (directory layout or hardcoded multilib more generally). We might want to find a more upstream-able solution at some point, but to unblock ourselves in the short term, just remove this support to prevent these build failures.

### DIFF
--- a/qualcomm-software/patches/llvm-project/0002-RISCV-Remove-hardcoded-RISC-V-multilib.patch
+++ b/qualcomm-software/patches/llvm-project/0002-RISCV-Remove-hardcoded-RISC-V-multilib.patch
@@ -1,0 +1,123 @@
+From 3dcb99366af084d1df9d0460f7817ec991698a91 Mon Sep 17 00:00:00 2001
+From: Jonathon Penix <jpenix@qti.qualcomm.com>
+Date: Fri, 6 Feb 2026 18:39:34 -0800
+Subject: [PATCH] [RISCV] Remove hardcoded RISC-V multilib
+
+When we build our embedded runtimes, we start with an empty directory as our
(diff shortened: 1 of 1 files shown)


jonathonpenix:
NB check-clang should fail here, I'll fix up the tests later if this works for what we need

~Edit: or not? I assumed there would have been some testing~ Edit2: Ah, there is, my local build was just missing the patch.

jonathonpenix:
Thinking about this a bit more/discussing this, the dummy multilib.yaml might be worthwhile to try as it might help with ex: explicit --sysroot usage into a variant.

jonathonpenix:
Closing, not using this workaround

PR #145, opened 2026-02-07
[cpullvm] Fix variant name

jonathonpenix:
There was an accidental mismatch here from when I renamed the fpic/nopic configs. Align the variant name here (used for build folders, xfail-ing tests, etc.) with the variant name in multilib.json for consistency.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/variants/riscv32imac_zba_zbb_ilp32.json
+++ b/qualcomm-software/embedded-multilib/json/variants/riscv32imac_zba_zbb_ilp32.json
@@ -2,7 +2,7 @@
     "args": {
         "common": {
             "TARGET_ARCH": "riscv32",
-            "VARIANT": "riscv32imac_zba_zbb_ilp32_fpic",
+            "VARIANT": "riscv32imac_zba_zbb_ilp32",
             "COMPILE_FLAGS": "-march=rv32imac_zba_zbb -mabi=ilp32 -fPIC",
             "ENABLE_EXCEPTIONS": "OFF",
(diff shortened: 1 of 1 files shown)


qualcomm-software/embedded-multilib/json/variants/riscv32imac_zba_zbb_ilp32.json
```diff
@@ -2,7 +2,7 @@
     "args": {
         "common": {
             "TARGET_ARCH": "riscv32",
-            "VARIANT": "riscv32imac_zba_zbb_ilp32_fpic",
+            "VARIANT": "riscv32imac_zba_zbb_ilp32",
             "COMPILE_FLAGS": "-march=rv32imac_zba_zbb -mabi=ilp32 -fPIC",
```
zhaoshiz, line 6: `-fPIC` is enabled here, IIRC that's the default setting, so our naming convention is no suffix for PIC and `nopic` for non-pic?

jonathonpenix: Yep!

There's one thing we need to fix for RISC-V before it happens, but our goal is to distribute _only_ fpic configs in the next release or so, which is also part of the motivation of not explicitly naming "fpic" configs.

zhaoshiz: Thanks for clarifying that!

zhaoshiz (approved):

PR #153, opened 2026-02-10
[cpullvm] Fix default external lit path for Windows

jonathonpenix:
llvm-lit usually has the '.py' suffix when building on Windows. Setting LLVM_DEFAULT_EXTERNAL_LIT to just `/path/to/llvm-lit` can lead to surprise test failures on Windows then as that path doesn't exist (but `/path/to/llvm-lit.py` does), so we ignore the specified path and fall back to other default paths.

Correct this by using the expected extension on Windows.

### DIFF
--- a/qualcomm-software/CMakeLists.txt
+++ b/qualcomm-software/CMakeLists.txt
@@ -463,7 +463,11 @@ add_dependencies(
 # Set LLVM_DEFAULT_EXTERNAL_LIT to the directory of clang
 # which was build in previous step. This path is not exported
 # by add_subdirectory of llvm project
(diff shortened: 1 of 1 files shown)


jonathonpenix:
Example of the failures observed: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/21851538159/job/63058998378?pr=150 ("C:\hostedtoolcache\windows\Python\3.13.11\x64\python3.exe: can't open file 'C:\\llvm-lit.py': [Errno 2] No such file or directory")

(And the clang failures further up in the log are unrelated)

navaneethshan (approved):

PR #158, opened 2026-02-11
Add a few more riscv64gc baremetal runtime variants

pzhengqc:
riscv64gc_zba_zbb_lp64d_nopic
riscv64gc_lp64_nopic
riscv64gc_zba_zbb_lp64_nopic

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -56,9 +56,27 @@
             "libraries_supported": "picolibc"
         },
         {
+            "variant": "riscv64gc_lp64_nopic",
+            "json": "riscv64gc_lp64_nopic.json",
+            "flags": "--target=riscv64-unknown-unknown-elf -march=rv64gc -mabi=lp64 -fno-pic -fno-exceptions",
+            "libraries_supported": "picolibc"
+        },
+        {
+            "variant": "riscv64gc_zba_zbb_lp64_nopic",
+            "json": "riscv64gc_zba_zbb_lp64_nopic.json",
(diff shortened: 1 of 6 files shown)


qualcomm-software/embedded-multilib/json/multilib.json
```diff
@@ -60,6 +60,24 @@
...
+            "flags": "--target=riscv64-unknown-unknown-elf -march=rv64gc_zba_zbb -mabi=lp64d -fno-pic -fno-exceptions",
+            "libraries_supported": "picolibc"
+        },
+        {
+            "variant": "riscv64gc_lp64_nopic",
+            "json": "riscv64gc_lp64_nopic.json",
+            "flags": "--target=riscv64-unknown-unknown-elf -march=rv64gc -mabi=lp64 -fno-pic -fno-exceptions",
+            "libraries_supported": "picolibc"
+        },
+        {
```
jonathonpenix, line 64: One minor nit: can you move riscv64gc_lp64_nopic and riscv64gc_zba_zbb_lp64_nopic here up after riscv64imac_lp64_nopic (but before riscv64gc_lp64d_nopic)? Just so the soft float/hard float rv64 configs are grouped together

jonathonpenix: (Not for these exact configs, but in general the order in the file can make a difference in what library is chosen, so I'm hoping we can keep similar configs grouped for if/when we do have to care about more subtle ordering cases).

pzhengqc: Makes sense. I've reordered the variants.

jonathonpenix (approved):
Thanks Pengxuan!

PR #159, opened 2026-02-11
[Multilib] Adding armv8_soft_neon variant

zhaoshiz:


qualcomm-software/embedded-multilib/json/variants/armv8_soft_neon.json
```diff
@@ -0,0 +1,32 @@
+{
+    "args": {
+        "common": {
+            "TARGET_ARCH": "armv8",
+            "VARIANT": "arvm8_soft_neon",
```
jonathonpenix, line 5: Minor nit/question, we do `armv7a` elsewhere, is it worth having the same distinction for armv8 (call the variant/files `armv8a_*`)? We don't have any armv8-m, but maybe worth just to be consistent? Also not sure if there's a standard convention.

I don't feel that strongly about it, will leave it up to you

qualcomm-software/embedded-multilib/json/multilib.json
```diff
@@ -35,6 +35,11 @@
             "json": "aarch64a_soft_nofp_pacret_bkey_bti.json",
             "flags": "--target=aarch64-unknown-none-elf -march=armv8.5-a -march=armvX+nofp -march=armvX+nosimd -mbranch-protection=pac-ret+leaf+b-key+bti -mabi=aapcs-soft"
         },
+        {
+            "variant": "armv8_soft_neon",
+            "json": "armv8_soft_neon.json",
+            "flags": "--target=thumbv8-unknown-none-eabi -mfpu=neon-fp-armv8 -fno-exceptions"
```
jonathonpenix: The `-fno-exceptions` here can be dropped

qualcomm-software/embedded-multilib/json/variants/armv8_soft_neon.json
zhaoshiz: we use armv8 without `-a` in musl: https://github.com/qualcomm/musl-embedded/blob/main/qualcomm-software/config/linux/arm/armv8_baremetal.sh. I'd like to keep it the same.

We can add that later.

qualcomm-software/embedded-multilib/json/multilib.json
zhaoshiz: fixed.

jonathonpenix (approved):

apazos (approved):

PR #161, opened 2026-02-12
[cpullvm] Install Linux libraries in `*-linux-musl[eabi]` directories

jonathonpenix:
Place our Linux libraries in `*-linux-musl` libraries to make it more clear how the sysroot is constructed (that is actually musl rather than ex: glibc).

Note though that we're still keeping the `*-linux-gnu` triples when building in keeping with what we've done in the past.


apazos (approved):
Thanks Jonathon, agree on changing the install dir only for now.

PR #164, opened 2026-02-12
Change variant name for riscv64imac_lp64_nopic

pzhengqc:
This seems to be the only exception. So change it to align with other variants.


jonathonpenix (approved):
Oops, missed that. Thanks!

jonathonpenix:
I think this is a limited/uncontroversial change so I'm going to go ahead and merge this

PR #166, opened 2026-02-12
[cpullvm][Multilib] Don't require matching against -fno-exceptions

jonathonpenix:
Currently, all of our configs have exception support disabled and people are required to pass -fno-exceptions to match any multilib. This becomes a bit unwieldy as this is required even for ex: C-only projects. We also don't particularly want to ship -fexceptions variants currently as a fallback or differ from community in terms of -fexceptions defaults.

So for now, just don't require matching -fno-exceptions to get a library without exception support.

This is a bit iffy in the sense that if we *do* need an -fexceptions variant, we'll have to rethink how we handle this situation. But, we've discussed this and decided to live with that possibility.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -3,100 +3,100 @@
         {
             "variant": "aarch64a",
             "json": "aarch64a.json",
-            "flags": "--target=aarch64-unknown-none-elf -fno-exceptions"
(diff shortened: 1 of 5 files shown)


jonathonpenix:
cc'ing @efriedma-quic since I can't add you as a reviewer

efriedma-quic (approved):
LGTM

apazos (approved):
Yes, agreed to disable exceptions and RTTI features since most of the baremetal images  do not need them and they also can cause code size increase.

PR #170, opened 2026-02-13
Cherry-pick patch for lld undefined behavior with SHT_GROUP.

efriedma-quic:
This is hitting an internal buildbot.  Not sure it'll actually hit cpullvm, but might as well bring it.


qualcomm-software/patches/llvm-project/0002-lld-Fix-undefined-behavior-with-misaligned-SHT_GROUP.patch
jonathonpenix, line 1: Sorry, can you also name this "0002-*"?

I think we (generally) sort these patches by name and apply in that order--it shouldn't matter right now for what we have, but just in (the unlikely) case anything else comes along/for consistency with other projects.

efriedma-quic: Done.

jonathonpenix (approved):
Thanks!

efriedma-quic:
Ping. (I don't have merge rights.)

jonathonpenix:
Oops, I just wasn't sure if we were making this conditional on whether we had a ubsan builder. Will merge now

PR #173, opened 2026-02-13
[Multilib] Add armv7m_soft_nofp variant

zhaoshiz:


qualcomm-software/embedded-multilib/json/multilib.json
```diff
@@ -45,6 +45,11 @@
             "json": "armv7a_soft_nofp.json",
             "flags": "--target=thumbv7-unknown-none-eabi -mfpu=none"
         },
+        {
+            "variant": "armv7m_soft_nofp",
+            "json": "armv7m_soft_nofp.json",
+            "flags": "--target=thumbv7m-unknown-none-eabi -mfpu=none -mno-unaligned-access"
```
jonathonpenix: Don't think we want the `-mno-unaligned-access` in multilib flags here

qualcomm-software/embedded-multilib/json/variants/armv7m_soft_nofp.json
```diff
@@ -0,0 +1,32 @@
+{
+    "args": {
+        "common": {
+            "TARGET_ARCH": "armv7m",
+            "VARIANT": "armv7m_soft_nofp",
+            "COMPILE_FLAGS": "-mfloat-abi=soft -march=armv7m -mfpu=none -mno-unaligned-access",
```
jonathonpenix: Hm, apparently this isn't PIC in musl-embedded. Do we want this to be PIC though? If, not, can we make this a no-pic config (mark it as such + make -fno-pic required in multilib)

jonathonpenix: FWIW, this is going to be implicitly built with -fPIC for musl-embedded in `embedded-runtimes/CMakeLists.txt`--at the time I didn't realize we had configs we cared about without -fPIC specified in the musl config scripts.

So, picolibc is the outlier here without changes (it'd be no-PIC, compiler-rt + musl would be PIC, etc.)

qualcomm-software/embedded-multilib/json/multilib.json
zhaoshiz: this flag is in ARM-embedded toolchain as well: https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/blob/main/arm-multilib/json/variants/armv7m_soft_nofp.json
If we tested picolibc for armv7m downstream without it I can move it to MUSL only

jonathonpenix: I don't mind having it in the compile flags for both, I'm just not sure we want to gate the multilib on it--we don't ship unaligned variants (and it seems like it should be safe to use the aligned version even if unaligned versions are supported?)

I think we did similar things for armv7a (`-mno-unaligned-access` in compile flags, but not require matching in multilib)

zhaoshiz: ok, I see, let me make the change.

qualcomm-software/embedded-multilib/json/variants/armv7m_soft_nofp.json
zhaoshiz: so, should we make this explicitly PIC including picolibc?

jonathonpenix: I guess that'd be my preference since at this point (since I'm not too sure who's using this, and to err towards being consistent with our other configs)

apazos: We discussed about having all the configs  with PIC / Atomics / TLS / Threading enabled.

zhaoshiz:
removed `-mno-unaligned-access` in multilib flag and added `-fPIC` in compiler flag

jonathonpenix (approved):
LGTM, thanks!

apazos (approved):

PR #150, opened 2026-02-08
[cpullvm][Github] Add Windows native runtimes builds

jonathonpenix:
QEMU tests are left disabled until we can sort out how to handle that
(both in terms of Windows, and in terms of Github-hosted runners).

PR #178, opened 2026-02-18
[cpullvm] Copy runtimes from the Linux x86_64 build in our main nightly

jonathonpenix:
Note that this currently only copies the *embedded* runtimes. Copying the
Linux runtimes requires a small change to the library copying script, which
I'd like to do separately.

There's also a bit of weirdness here in that, if ex: an eld lit test fails
in the Linux x86_64 builder, we won't start builds in any of the other hosts.
That probably isn't great in the longer term and we can probably fix this,
but for now let's start simple and we can fine tune from there. We need the
Linux x86_64 builder to be clean anyways so hopefully we don't hit issues
with this that often.


jonathonpenix:
~Stacking this on top of https://github.com/qualcomm/cpullvm-toolchain/pull/177 and https://github.com/qualcomm/cpullvm-toolchain/pull/150 as 177 should be needed to test on Windows and 150 makes it a bit easier to see where we end up in terms of what scripts are needed (the number is going down!)~

~Ignore the first two commits when looking at this PR.~

Those are now merged so never mind that part!

jonathonpenix:
Successful build of the nightlies: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/22170959385?pr=178

navaneethshan (approved):

jonathonpenix:
This should finally be good to go:
* Fixed accidentally re-uploading the x86 Linux tarball along with the Windows and AArch64 Linux builds
* There was some issue with files not getting the right `.zip` extension. The timing was unfortunate, but turns out this was a Github-side issue, not an issue with this patch.

Fixed nightly run: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/22241566207

apazos (approved):

PR #179, opened 2026-02-18
Add riscv64 SCS baremetal runtime variants

pzhengqc:
Testing for picolibc is disabled due to lack of runtime support.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -122,6 +122,12 @@
             "libraries_supported": "picolibc"
         },
         {
+            "variant": "riscv64imac_lp64_scs_nopic",
+            "json": "riscv64imac_lp64_scs_nopic.json",
+            "flags": "--target=riscv64-unknown-unknown-elf -march=rv64imac -mabi=lp64 -fsanitize=shadow-call-stack -fno-pic",
+            "libraries_supported": "picolibc"
+        },
+        {
             "variant": "riscv64gc_lp64_nopic",
             "json": "riscv64gc_lp64_nopic.json",
             "flags": "--target=riscv64-unknown-unknown-elf -march=rv64gc -mabi=lp64 -fno-pic",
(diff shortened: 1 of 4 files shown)


pzhengqc:
The test failure is due to -fmultilib-flag not supported by RISC-V and we need https://github.com/qualcomm/cpullvm-toolchain/pull/182 to unblock this patch.

pzhengqc:
I verified this patch together with https://github.com/qualcomm/cpullvm-toolchain/pull/182. Refer to https://github.com/qualcomm/cpullvm-toolchain/pull/181 for the verification results.

jonathonpenix:
I think since `-fsanitize...` is a valid compiler flag, we might want to add support to recognize that flag in multilib (in `getMultilibFlags` in ToolChain.cpp) rather than doing it as a custom flag.

(But, we'll need the custom flag support for ex: threading in picolibc so it is good that we have that still 😄)

pzhengqc:
> I think since `-fsanitize...` is a valid compiler flag, we might want to add support to recognize that flag in multilib (in `getMultilibFlags` in ToolChain.cpp) rather than doing it as a custom flag.
> 
> (But, we'll need the custom flag support for ex: threading in picolibc so it is good that we have that still 😄)

Agreed. Using multilib custom flag is just to avoid introducing changes to the driver, but if that's acceptable it's a cleaner solution I think. Thanks for bring up the alternative, @jonathonpenix.

jonathonpenix:
Driver changes are ok as long as we also upstream it I think 😄 (and, I think this is a reasonable enough change to upstream)

pzhengqc:
Here's the patch which exposes the sanitizer flags as multilib flags: https://github.com/qualcomm/cpullvm-toolchain/pull/185. Combined verification is successful (https://github.com/qualcomm/cpullvm-toolchain/pull/181).

qualcomm-software/embedded-multilib/json/variants/riscv64gc_scs_lp64_nopic.json
```diff
@@ -0,0 +1,26 @@
+{
+    "args": {
+        "common": {
+            "TARGET_ARCH": "riscv64",
+            "VARIANT": "riscv64gc_scs_lp64_nopic",
```
jonathonpenix: As a minor nit on naming, can we keep the extensions + abi together? So ex: `riscv64gc_lp64_scs_nopic` instead

Same for the imac config, etc.

jonathonpenix (commented):
Minor nit on naming of variants, but otherwise LGTM

Thanks Pengxuan!

qualcomm-software/embedded-multilib/json/variants/riscv64gc_scs_lp64_nopic.json
pzhengqc: Thanks, @jonathonpenix. I've updated all the variant names accordingly.

jonathonpenix (approved):
LGTM, thanks!

apazos (approved):
Thanks @pzhengqc for getting the driver changes upstreamed: https://github.com/llvm/llvm-project/pull/182350
https://github.com/llvm/llvm-project/pull/182220

PR #180, opened 2026-02-18
Add riscv32ima_xqci baremetal runtime variant

svs-quic:


svs-quic:
There is one known test failure in the picolibc tests. @RamNalamothu had disabled this test downstream for Xqci. I dont think this has been fixed in the latest QEMU as well. What do we for CPULLVM?

2026-02-18T08:24:19.4599715Z Failed Tests (1):
2026-02-18T08:24:19.4599960Z   picolibc-rv32ima_xqci_ilp32_nopic :: test-except.test

jonathonpenix:
If it is just this one test failing, please add it to `qualcomm-software/embedded-runtimes/test-support/xfails.py` (there should be some other xfails in there to look at as examples) and we can leave the rest enabled.

qualcomm-software/embedded-multilib/json/multilib.json
```diff
@@ -109,6 +109,12 @@
             "flags": "--target=riscv32-unknown-unknown-elf -march=rv32imafc_zcb_zcmp_zba_zbb -mabi=ilp32f",
             "libraries_supported": "picolibc"
         },
+        {
+            "variant": "rv32ima_xqci_ilp32_nopic",
+            "json": "rv32ima_xqci_ilp32_nopic.json",
+            "flags": "--target=riscv32-unknown-unknown-elf -march=rv32ima_zba_zbb_zbs_zca_zcb_zilsd_zclsd_xqcia_xqciac_xqcibi_xqcibm_xqcicli_xqcicm_xqcics_xqcicsr_xqciint_xqciio_xqcilb_xqcili_xqcilia_xqcilo_xqcilsm_xqcisim_xqcisls_xqcisync_xqccmp -mabi=ilp32 -fno-pic",
```
jonathonpenix, line 115: I think this is fine for now since we don't have Sam's patch to do the reverse implication thing for `xqci` (listing all the extensions adds xqci to the normalized `-march` string).

Once we go back to mainline and _do_ have that patch, I hope we can just use `xqci` instead (both for baremetal and Linux variants).

qualcomm-software/test/multilib/riscv32.test
```diff
@@ -39,6 +39,10 @@
 # CHECK-IMAFC-ZCB-ZCMP-ZBA-ZBB-ILP32F-PIC: riscv32-unknown-elf/riscv32imafc_zcb_zcmp_zba_zbb_ilp32f{{$}}
 # CHECK-IMAFC-ZCB-ZCMP-ZBA-ZBB-ILP32F-PIC-EMPTY:
 
+# RUN: %clang -print-multi-directory --target=riscv32-unknown-elf -march=rv32ima_zba_zbb_zbs_zca_zcb_zilsd_zclsd_xqcia_xqciac_xqcibi_xqcibm_xqcicli_xqcicm_xqcics_xqcicsr_xqciint_xqciio_xqcilb_xqcili_xqcilia_xqcilo_xqcilsm_xqcisim_xqcisls_xqcisync_xqccmp -mabi=ilp32 -fno-pic 2>&1 | FileCheck %s --check-prefix=CHECK-IMA-XQCI-ILP32-NO-PIC
```
jonathonpenix, line 42: Can you please also add a test that if someone just uses `xqci` (instead of listing all the xqci extensions) they also get the right library?

(I think this should work without any extra changes in the CMake/multilib code)

qualcomm-software/embedded-runtimes/CMakeLists.txt
```diff
@@ -176,7 +176,11 @@ if(ENABLE_LIBC_TESTS OR ENABLE_COMPILER_RT_TESTS OR ENABLE_LIBCXX_TESTS)
         if(TARGET_ARCH MATCHES "^aarch64")
             find_program(QEMU_EXECUTABLE qemu-system-aarch64 REQUIRED)
         elseif(TARGET_ARCH MATCHES "^riscv32")
-            find_program(QEMU_EXECUTABLE qemu-system-riscv32 REQUIRED)
+            if(VARIANT MATCHES "^rv32ima_xqci_ilp32")
+                find_program(QEMU_EXECUTABLE qemu-system-riscv32-xqci REQUIRED)
```
jonathonpenix, line 180: I'm a bit unsure about what to do here--I think this being required is going to be a nuisance for the majority of people who build CPULLVM given that this qemu is not a standard thing.

At minimum, we need to document what `qemu-system-riscv32-xqci` is and how to build/source it (probably in https://github.com/qualcomm/cpullvm-toolchain/blob/qualcomm-software/qualcomm-software/docs/building.md#prerequisites). This can be a separate PR, or in this one--either way works for me.

Beyond that, would it make sense to make this _not_ required and:
* if found, we can just continue as normal
* if _not_ found, maybe we just produce a warning and disable the tests for this variant? Somewhat similar to how ex: `ENABLE_QEMU_TESTING` works in 'embedded-multilib' [here](https://github.com/qualcomm/cpullvm-toolchain/blob/2b1e2073c2931363c5f302a1dd5da8c16c49429c/qualcomm-software/embedded-multilib/CMakeLists.txt#L197-L201)

My guess though is this warning will almost certainly get lost in the build logs so will probably hide issues for someone at some point who _does_ want to test xqci. Which, isn't great.

But, requiring everyone to do one of a) build our QEMU from source b) disable QEMU testing or c) disable just the xqci variant to get a successful build of CPULLVM seems like a large-ish barrier.

qualcomm-software/embedded-multilib/json/variants/riscv32ima_xqci_ilp32_nopic.json
```diff
@@ -0,0 +1,26 @@
...
+    "args": {
+        "common": {
+            "TARGET_ARCH": "riscv32",
+            "VARIANT": "rv32ima_xqci_ilp32_nopic",
+            "COMPILE_FLAGS": "-march=rv32ima_zba_zbb_zbs_zca_zcb_zilsd_zclsd_xqcia_xqciac_xqcibi_xqcibm_xqcicli_xqcicm_xqcics_xqcicsr_xqciint_xqciio_xqcilb_xqcili_xqcilia_xqcilo_xqcilsm_xqcisim_xqcisls_xqcisync_xqccmp -mabi=ilp32 -fno-pic",
+            "ENABLE_EXCEPTIONS": "OFF",
+            "ENABLE_RTTI": "OFF",
+            "TEST_EXECUTOR": "qemu",
+            "QEMU_MACHINE": "virt",
+            "QEMU_CPU": "rv32,zcd=false,d=false,zcb=true,zcmp=false,zilsd=true,zclsd=true,f=false,zfa=false,xqci=true,smrnmi=true,xqccmp=true",
```
jonathonpenix, line 11: For this, don't we have an xqci-specific QEMU CPU we can set?

apazos: Is it possible to selectively disable building specific runtime variants? While we can already skip building ARM or RISC-V runtimes, can we make exclusions at a more granular level? I agree, it will be painful for customers if this breaks their build. For now I suggest if QEMU not found, we can produce a warning and skip tests.

Regarding QEMU CPU name support, we can address it later. This should not delay merging the patch, we can explicitly enable the sub-extensions when invoking QEMU for now.

I agree, we should work on a PR to provide proper documentation on building QEMU for Xqci.

jonathonpenix: > Is it possible to selectively disable building specific runtime variants? While we can already skip building ARM or RISC-V runtimes, can we make exclusions at a more granular level?

Right now, you either build everything, or you list out exactly the individual variants you want (there are no larger "arm" or "riscv" groupings). So, ex: excluding _just_ the xqci config is somewhat painful.

The rest sounds good to me.

qualcomm-software/embedded-runtimes/CMakeLists.txt
lenary: I agree that the lack of this version of qemu should cause a warning and disable tests.

RamNalamothu:
> There is one known test failure in the picolibc tests. @RamNalamothu had disabled this test downstream for Xqci. I dont think this has been fixed in the latest QEMU as well. What do we for CPULLVM?
> 
> 2026-02-18T08:24:19.4599715Z Failed Tests (1): 2026-02-18T08:24:19.4599960Z picolibc-rv32ima_xqci_ilp32_nopic :: test-except.test

Yes, for xqci, this is the only test failing in the downstream and the latest QEMU didn't fix this issue.

qualcomm-software/embedded-runtimes/CMakeLists.txt
RamNalamothu: The specific naming is to disambiguate which QEMU to use for xqci vs non-xqci RISC-V configs if both QEMU binaries were to co-exist in the same environment.

Short term, if there is a way to explicitly point to the QEMU that needs to be used for a specific picolibc config for testing, the two QEMU binaries can be installed at different paths, and we can skip the special naming for the binary itself.

Long term, it is best if we could use the xqci specific QEMU for non-xqci configs testing as well so that there will be only one QEMU for everything i.e. for all of xqci and non-xqci RISC-V configs.

qualcomm-software/test/multilib/riscv32.test
```diff
@@ -39,6 +39,14 @@
...
+# CHECK-IMA-XQCI-ILP32-NO-PIC: riscv32-unknown-elf/riscv32ima_xqci_ilp32{{$}}
```
jonathonpenix:
```suggestion
# CHECK-IMA-XQCI-ILP32-NO-PIC: riscv32-unknown-elf/riscv32ima_xqci_ilp32_nopic{{$}}
```

qualcomm-software/test/multilib/riscv32.test
```diff
@@ -39,6 +39,14 @@
...
+# CHECK-IMA-XQCI-ILP32-NO-PIC-EMPTY:
+
+# RUN: %clang -print-multi-directory --target=riscv32-unknown-elf -march=rv32ima_zba_zbb_zbs_zca_zcb_zilsd_zclsd_xqci_xqccmp -mabi=ilp32 -fno-pic 2>&1 | FileCheck %s --check-prefix=CHECK-IMA-XQCI-ILP32-NO-PIC
+# CHECK-IMA-XQCI-ILP32-NO-PIC: riscv32-unknown-elf/riscv32ima_xqci_ilp32{{$}}
```
jonathonpenix: Think it might be clearer to just share the same CHECK thing as above rather than redefining(?) it.

qualcomm-software/embedded-multilib/json/multilib.json
```diff
@@ -109,6 +109,12 @@
             "flags": "--target=riscv32-unknown-unknown-elf -march=rv32imafc_zcb_zcmp_zba_zbb -mabi=ilp32f",
             "libraries_supported": "picolibc"
         },
+        {
+            "variant": "rv32ima_xqci_ilp32_nopic",
```
jonathonpenix:
```suggestion
            "variant": "riscv32ima_xqci_ilp32_nopic",
```
Just to keep consistent with our other variants

(Although thinking about it now, our Linux and baremetal variants are weirdly misaligned as-is, oops. But, I don't think we can fix that for this release at this point.)

qualcomm-software/embedded-multilib/json/multilib.json
```diff
@@ -109,6 +109,12 @@
...
+            "json": "rv32ima_xqci_ilp32_nopic.json",
```
jonathonpenix: Same rv->riscv

qualcomm-software/embedded-runtimes/CMakeLists.txt
```diff
@@ -176,7 +176,11 @@ if(ENABLE_LIBC_TESTS OR ENABLE_COMPILER_RT_TESTS OR ENABLE_LIBCXX_TESTS)
         if(TARGET_ARCH MATCHES "^aarch64")
             find_program(QEMU_EXECUTABLE qemu-system-aarch64 REQUIRED)
         elseif(TARGET_ARCH MATCHES "^riscv32")
-            find_program(QEMU_EXECUTABLE qemu-system-riscv32 REQUIRED)
+            if(VARIANT MATCHES "^rv32ima_xqci_ilp32")
```
jonathonpenix: rv -> riscv

qualcomm-software/embedded-runtimes/test-support/xfails.py
```diff
@@ -150,6 +150,18 @@ def check_r52_warning():
...
         ),
+        XFail(
+            name="picolibc_rv32ima_xqci",
+            testnames=[
+                "test-except"
+            ],
+            result=NewResult.EXCLUDE,
+            project="picolibc",
+            variants=[
+                "riscv32ima_xqci_ilp32_nopic"
```
jonathonpenix, line 161: The test exclude/xfail issues might be due to the variant name here--we define the variant as `rv32ima*` but list `riscv32ima*` here. Once this is made consistent, this _hopefully_ will work as expected.

I'll try locally and report back.

Edit: I didn't try this exact test (I don't have the xqci qemu locally), but when trying to exclude other tests, it looks like once we get the variant name mismatch sorted out, we can just go back to the full name (test-except.test) and things should be excluded correctly

svs-quic: Thanks Jonathon. I've made these changes and the build finally passes.

qualcomm-software/embedded-runtimes/CMakeLists.txt
svs-quic: I would like to do the documentation and adding of the warning in a follow-up patch if that's alright.

qualcomm-software/embedded-multilib/json/variants/riscv32ima_xqci_ilp32_nopic.json
```diff
@@ -0,0 +1,26 @@
...
+            "QEMU_PARAMS": "-bios none",
+            "FLASH_ADDRESS": "0x80000000",
+            "FLASH_SIZE": "0x00400000",
+            "RAM_ADDRESS": "0x80400000",
+            "RAM_SIZE": "0x00200000",
+            "LIBRARY_BUILD_TYPE": "minsizerelease"
+        },
+        "picolibc": {
+            "ENABLE_CXX_LIBS": "ON",
```
jonathonpenix, line 20: Not requesting changes here, but FWIW I think we have this turned off downstream. I'm fine/happy if we leave them on (seems like they build fine) in case anyone wants to experiment with them. But, not sure what folks expect/if this was something folks discussed.

qualcomm-software/embedded-runtimes/CMakeLists.txt
jonathonpenix: Sure, that works for me

jonathonpenix (approved):
Left one misc. comment about whether we wanted c++ libs, but I don't mind either way.

I also am fine if we document QEMU + warn/disable tests if QEMU not found in follow-up patches.

So, LGTM, thanks Sudharsan!

qualcomm-software/embedded-multilib/json/variants/riscv32ima_xqci_ilp32_nopic.json
svs-quic: I don't think we need them but given that things build fine let us merge this as is and I can remove them later if we run into issues.

jonathonpenix: Discussed offline, and will go ahead with this for now. We can revisit later.

PR #182, opened 2026-02-18
Extend support for -fmultilib-flag to all targets

pzhengqc:
Support for the -fmultilib-flag option has been added to AArch64/ARM. This patch extends the support of the flag to all targets.

### DIFF
--- a/qualcomm-software/patches/llvm-project/0003-Multilib-Extend-support-for-fmultilib-flag-to-all-ta.patch
+++ b/qualcomm-software/patches/llvm-project/0003-Multilib-Extend-support-for-fmultilib-flag-to-all-ta.patch
@@ -1,0 +1,61 @@
+From 6c18d648ac2f578998f8dda664ffde9cc1667c0c Mon Sep 17 00:00:00 2001
+From: Pengxuan Zheng <pzheng@qti.qualcomm.com>
+Date: Wed, 18 Feb 2026 21:11:43 -0800
+Subject: [PATCH] [Multilib] Extend support for -fmultilib-flag to all targets
+
+Support for the `-fmultilib-flag` option has been added to AArch64/ARM. This
+patch extends the support of the flag to all targets.
+---
+ clang/lib/Driver/ToolChain.cpp                  | 6 ++----
+ clang/test/Driver/print-multi-selection-flags.c | 2 ++
+ 2 files changed, 4 insertions(+), 4 deletions(-)
+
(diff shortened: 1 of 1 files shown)


qualcomm-software/patches/llvm-project/0002-Multilib-Add-support-for-fmultilib-flag-to-RISC-V.patch
```diff
@@ -0,0 +1,27 @@
...
+diff --git a/clang/lib/Driver/ToolChain.cpp b/clang/lib/Driver/ToolChain.cpp
+index 77a2c73f0d44..87912ab13a59 100644
+--- a/clang/lib/Driver/ToolChain.cpp
++++ b/clang/lib/Driver/ToolChain.cpp
+@@ -334,6 +334,8 @@ static void getRISCVMultilibFlags(const Driver &D, const llvm::Triple &Triple,
+     Result.push_back("-march=" + (*ISAInfo)->toString());
+ 
+   Result.push_back(("-mabi=" + riscv::getRISCVABI(Args, Triple)).str());
++
++  processMultilibCustomFlags(Result, Args);
```
jonathonpenix: Is it worth just adding this to `getMultilibFlags` instead (and removing the usages from the Arm/AArch64 functions)?

I'm not really sure why this was only added for Arm/AArch64 originally upstream, but it doesn't really seem like this is a target-specific flag.

jonathonpenix (approved):
Left one question, but that is more for when this is upstreamed--I don't have an issue if we merge this as-is.

Thanks Pengxuan!

qualcomm-software/patches/llvm-project/0002-Multilib-Add-support-for-fmultilib-flag-to-RISC-V.patch
pzhengqc: It doesn't seem like a target-specific flag to me either. I can try adding this to `getMultilibFlags` when upstreaming the patch and see if there's any objection.

pzhengqc:
The patch has been sent to LLVM upstream for review: https://github.com/llvm/llvm-project/pull/182220

jonathonpenix (approved):

pzhengqc:
Updated the patch based on reviewer feedback from LLVM upstream.

apazos (approved):

PR #186, opened 2026-02-20
workflows/issue-release-workflow: Add an environment in order to limi…

navaneethshan:
…t secret access (#179585)

I've created a new environment called `main-branch-only` which is meant to be used for running the GitHub Actions based automations for the project. By using an environment, we can move some of the secrets we use from the repository configuration and into the environment configuration, which means they will only be accessible for jobs that run on the branch we specify, which in this case is the main branch.

This will prevent people from using user branches as a way to gain access to secret values (which they can do now very easily by creating a workflow that dumps the secrets). Even though someone could still do this by pushing to main, user branches are hard to monitor and they can be easily deleted after dumping the secrets leaving no trace.

If someone where to do this on the main branch it would be much more visible and impossible to cover up.

I would like to apply this to more workflows, but I'm starting here so I can test it and make sure it works the way I expect.

(cherry picked from commit 4e21141bae0dac9eadd28fe230ffe213cd156af6)

### DIFF
--- a/.github/workflows/issue-release-workflow.yml
+++ b/.github/workflows/issue-release-workflow.yml
@@ -1,0 +1,70 @@
+# This contains the workflow definitions that allow users to test backports
+# to the release branch using comments on issues.
+#
+# /cherry-pick <commit> <...>
+#
+# This comment will attempt to cherry-pick the given commits to the latest
+# release branch (release/Y.x) and if successful, push the result to a branch
+# on github.
+#
+# /branch <owner>/<repo>/<branch>
+#
+# This comment will create a pull request from <branch> to the latest release
+# branch.
+
+name: Issue Release Workflow
+
+permissions:
+  contents: read
+
+on:
+  issue_comment:
+    types:
+      - created
+      - edited
+  issues:
+    types:
+      - opened
+
+env:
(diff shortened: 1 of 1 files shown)


navaneethshan:
I requested for the approval for the token with pull-request and write permission, but this change can be reviewed/merged while we wait.

.github/workflows/issue-release-workflow.yml
```diff
@@ -0,0 +1,69 @@
...
+          # GitHub stores the token used for checkout and uses it for pushes
+          # too, but we want to use a different token for pushing, so we need
+          # to disable persist-credentials here.
+          persist-credentials: false
+          fetch-depth: 0
+
+      - name: Setup Environment
+        run: |
+          pip install --require-hashes -r ./llvm/utils/git/requirements.txt
+          ./llvm/utils/git/github-automation.py --token ${{ github.token }} setup-llvmbot-git
```
jonathonpenix: We might want to do something different here? Not sure if it will be confusing/cause problems if we use the llvm.org emails.

navaneethshan: Good point, please check now.

jonathonpenix (approved):
LGTM, thanks!

Once this is in I think we have to mess with the milestones to test--hopefully that will be the last thing!

PR #188, opened 2026-02-21
[cpullvm] Generalize clang xfail generation to work with other llvm projects

jonathonpenix:
This lets us handle other projects in LLVM (lldb is the main motivating force
here) similar to what we can do for clang.

Not all projects have an equivalent to CLANG_TEST_EXTRA_ARGS though so we use
LLVM_LIT_ARGS instead to pass in the xfails list. But, we need to set
LLVM_LIT_ARGS before we add LLVM as a subproject, so we can't just rely on the
defaults like we do now. (See ex: how our compiler-rt xfails work.) I don't
think this is a huge deal, but maybe it'll cause some surprises.

### DIFF
--- a/qualcomm-software/CMakeLists.txt
+++ b/qualcomm-software/CMakeLists.txt
@@ -195,6 +195,7 @@ set(LLVM_BUILD_RUNTIME OFF CACHE BOOL "")
 set(LIBCLANG_BUILD_STATIC ON CACHE BOOL "")
 set(LLVM_POLLY_LINK_INTO_TOOLS ON CACHE BOOL "")
 set(CLANG_DEFAULT_LINKER lld CACHE STRING "")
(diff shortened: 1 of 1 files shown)


jonathonpenix:
Sorry, never mind this actually--we probably want something like this eventually, but this requires more work to decide how to mark dependencies between the xfails files and arbitrary `check-` targets since with this, _all_ `check-` need these files.

jonathonpenix:
Going to close this for now since I don't think we have a particular need at the moment (lldb tests are disabled/we're trying to get them clean, we've generally just been dealing with eld failures as they come up).

In case it is helpful in the future, I actually think this is what ATfE _used_ to do before narrowing the scope to be clang only--so if this ever needs to be revived, should check that first.

PR #196, opened 2026-02-24
[cpullvm] Enable copying Linux libraries from existing distribution files

jonathonpenix:
This adds an off-by-default flag `--include-linux-libraries` to copy_target_libraries.py to accomplish this.

Turn this new flag on in our nightlies that copy runtimes too.

### DIFF
--- a/qualcomm-software/CMakeLists.txt
+++ b/qualcomm-software/CMakeLists.txt
@@ -134,8 +134,8 @@ option(
     "Include additional library variants for use on Linux targets"
     OFF
 )
-if(ENABLE_LINUX_LIBRARIES AND WIN32)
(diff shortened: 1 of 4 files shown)


jonathonpenix:
Note that the commit adding the `FIXME` change needs to be dropped before merging, but waiting to get a build of the nightly config with this change.

jonathonpenix:
Successful nightly build with these changes: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/22375493806?pr=196

apazos (approved):

jonathonpenix:
Failures are unrelated (and I just fixed up a comment and it was passing before that) so I'm going to merge this.

PR #198, opened 2026-02-25
[ATfE] Parallelize building compiler-rt profile library (#668)

jonathonpenix:
Configuring and building the compiler-rt profile library for an individual variant does not take much time, but done one after another in series it becomes significant. This patch allows the library to be built in parallel like compiler-rt itself, which is significantly faster.

(cherry picked from commit d2ed97d1154c703b9c9d928d9f3de2bae56ac73b)


jonathonpenix:
Note that this is just a cherry-pick from ATfE: https://github.com/arm/arm-toolchain/commit/d2ed97d1154c703b9c9d928d9f3de2bae56ac73b

apazos (approved):

PR #202, opened 2026-02-26
Add compiler-rt build for Windows targets.

efriedma-quic:
compiler-rt isn't usually necessary on Windows, but there are some libraries with more obscure uses: the builtins library has SME support routines, the profile library is useful for PGO and coverage, the ubsan library is useful for enabling ubsan.

Builds both x64 and arm64 libraries... not sure we really need the x64 versions, but the build steps are basically the same.

(Only works on Windows host for obvious reasons.)

### DIFF
--- a/qualcomm-software/CMakeLists.txt
+++ b/qualcomm-software/CMakeLists.txt
@@ -134,9 +134,18 @@ option(
     "Include additional library variants for use on Linux targets"
     OFF
 )
+option(
+    ENABLE_WINDOWS_LIBRARIES
+    "Include additional library variants for use on Windows targets"
+    "${CMAKE_HOST_WIN32}"
+)
 if((ENABLE_LINUX_LIBRARIES AND WIN32) AND NOT PREBUILT_TARGET_LIBRARIES)
     message(FATAL_ERROR "Building Linux libraries from source is not supported on Windows")
 endif()
+if(ENABLE_WINDOWS_LIBRARIES AND NOT CMAKE_HOST_WIN32)
+    message(FATAL_ERROR "Building Windows libraries from source is not supported on Linux")
+endif()
+
 option(LLVM_TOOLCHAIN_LIBRARY_OVERLAY_INSTALL
     "Make cpack build an overlay package that can be unpacked over the main toolchain to install a secondary set of libraries based on musl-embedded."
     ${overlay_install_default})
@@ -651,6 +660,38 @@ if(NOT PREBUILT_TARGET_LIBRARIES)
     endif()
 endif()
 
+if(ENABLE_WINDOWS_LIBRARIES)
+    set(windows_library_prefix ${CMAKE_BINARY_DIR}/windows-library-builds)
+    ExternalProject_Add(
+        windows-libraries
+        STAMP_DIR ${windows_library_prefix}/stamp
+        BINARY_DIR ${windows_library_prefix}/build
+        TMP_DIR ${windows_library_prefix}/tmp
+        SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/scripts
+        INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/llvm/windows-libraries
+        DEPENDS ${lib_tool_dependencies}
+        CONFIGURE_COMMAND ""
+        BUILD_COMMAND
+            powershell -ExecutionPolicy Bypass -File "<SOURCE_DIR>/build_windows_runtimes.ps1"
+                -ToolsPath ${CMAKE_CURRENT_BINARY_DIR}/llvm/bin
+                -BaseBuildDir <BINARY_DIR>
+                -BaseInstallDir <INSTALL_DIR>
+                -LLVMSrcDir ${llvmproject_src_dir}
+        INSTALL_COMMAND ""
+        USES_TERMINAL_CONFIGURE TRUE
+        USES_TERMINAL_BUILD TRUE
+        USES_TERMINAL_INSTALL TRUE
+        STEP_TARGETS build install
+        LOG_BUILD TRUE
+        BUILD_ALWAYS TRUE
+    )
+
+    add_dependencies(
+        llvm-toolchain-runtimes
+        windows-libraries-install
+    )
+endif()
+
 FILE(WRITE "${CMAKE_BINARY_DIR}/llvm/${TARGET_CFI_DIR}/cfi_ignorelist.txt" "")
 install(
     FILES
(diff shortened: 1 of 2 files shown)


jonathonpenix:
>Currently installs into "windows-libraries/" in the install directory, similar to what we do for "linux-libraries/"; not sure if we should try to install to "lib\clang\22" like upstream releases.

The situation is a bit odd, but Linux libraries (and kind of the embedded libraries too) are two-stage-ish thing right now: the script "installs" the libraries into the `linux-libraries` folder in the build folder, but when we do the "real" install, cmake knows how to copy from `linux-libraries` into `lib/clang/22`. See the snippet [here](https://github.com/qualcomm/cpullvm-toolchain/blob/3507be419945eaddbed7a5d3d58b31096c9fc945/qualcomm-software/CMakeLists.txt#L668-L682)

I think we need an equivalent second step for the Windows libs here

(Edit: I might be misreading this. I agree that maybe we should have the script install directly into the correct locations in the _build_ folder too rather than `linux-libraries`/`windows-libraries`, but I think the second step for the install is necessary either way. The intermediate `linux-libraries` is originally just from it being a bit awkward to decide where to install things in the script itself and our library layout isn't known to clang anyway. But, it does make using a 'build' only toolchain a bit annoying)

qualcomm-software/scripts/build_windows_runtimes.ps1
```diff
@@ -0,0 +1,103 @@
...
+      -DCMAKE_INSTALL_PREFIX="$BaseInstallDir" `
+      -DCMAKE_BUILD_TYPE="Release" `
+      -DCMAKE_C_COMPILER="clang-cl" `
+      -DCMAKE_CXX_COMPILER="clang-cl" `
+      -DCMAKE_SYSTEM_NAME="Windows" `
+      -DCOMPILER_RT_DEFAULT_TARGET_ONLY=ON `
+      -DCMAKE_ASM_COMPILER_TARGET="$VariantTarget" `
+      -DCMAKE_C_COMPILER_TARGET="$VariantTarget" `
+      -DCMAKE_CXX_COMPILER_TARGET="$VariantTarget" `
+      -DCMAKE_ASM_FLAGS="${LIB_BUILD_FLAGS}" `
```
jonathonpenix: Is LIB_BUILD_FLAGS expected to be set to anything here?

qualcomm-software/CMakeLists.txt
```diff
@@ -134,6 +134,11 @@ option(
     "Include additional library variants for use on Linux targets"
     OFF
 )
+option(
+    ENABLE_WINDOWS_LIBRARIES
+    "Include additional library variants for use on Windows targets"
+    "$CMAKE_HOST_WIN32"
```
jonathonpenix: I'm a bit surprised cmake isn't grumpy about this, but my guess is `{}` are needed here for this to work as expected

Also, might be worth throwing an error early if someone tries to enable this on Linux? Up to you

qualcomm-software/scripts/build_windows_runtimes.ps1
```diff
@@ -0,0 +1,103 @@
...
+      -DCMAKE_C_FLAGS="${LIB_BUILD_FLAGS}" `
+      -DCMAKE_CXX_FLAGS="${LIB_BUILD_FLAGS}" `
+      -DLLVM_ENABLE_PER_TARGET_RUNTIME_DIR=ON `
+      -DCOMPILER_RT_CXX_LIBRARY="libcxx" `
```
jonathonpenix: Would we need to start building/shipping libc++ on windows (or does setting this include a build?) for people to be able to use the libraries if we're building against libc++ here?

I don't really know how this works offhand, so maybe it isn't an issue.

qualcomm-software/scripts/build_windows_runtimes.ps1
```diff
@@ -0,0 +1,103 @@
...
+      -DCOMPILER_RT_BUILD_BUILTINS=ON `
+      -DCOMPILER_RT_BUILD_LIBFUZZER=OFF `
+      -DCOMPILER_RT_SANITIZERS_TO_BUILD="" `
```
jonathonpenix: Just double checking, you mentioned ubsan in the commit message, but we aren't expecting it to be built/distribute it, correct?

qualcomm-software/scripts/build_windows_runtimes.ps1
efriedma-quic: This was copied from the Linux script; I guess I don't currently have any flags I need.

qualcomm-software/scripts/build_windows_runtimes.ps1
efriedma-quic: I can drop this, I think; we aren't currently building any libraries that actually require a C++ library.

qualcomm-software/CMakeLists.txt
efriedma-quic: The rules are shell interpolation rules: it works with or without braces, but some cases are ambiguous if you don't use them.

I can add an error, sure.

efriedma-quic (commented):
> I think we need an equivalent second step for the Windows libs here

As far as I can tell, the second step is necessary on the Linux side because the libraries are rearranged after they're installed.  If we're just going to do a straight copy of everything, I don't see the point to using a temporary directory.

qualcomm-software/CMakeLists.txt
jonathonpenix: >The rules are shell interpolation rules: it works with or without braces, but some cases are ambiguous if you don't use them.

I'm not following this--are you saying this is getting preprocessed by something before cmake gets to it? Otherwise, these aren't the same in cmake as far as I know (ex: https://godbolt.org/z/5a37vM91j).

I might be misunderstanding something though?

jonathonpenix:
The re-arranging is specific to Linux (and I agree/don't mind with doing away with that part) but the extra copy itself from build to install isn't specific to the Linux libraries here--cmake otherwise doesn't copy that folder as part of the install (even if the shell/powershell scripts place the libraries to the correct lib/resource dir in the build folder). (And that isn't specific to us here, same happens with upstream LLVM if you create your own `lib` etc. in the resource dir)

I guess the scripts could _maybe_ just place things in the final install location directly, but that seems like it has its own set of problems that aren't necessarily better.

qualcomm-software/CMakeLists.txt
efriedma-quic: Hmm.  Maybe I'm just confused.

efriedma-quic:
> but the extra copy itself from build to install isn't specific to the Linux libraries here--cmake otherwise doesn't copy that folder as part of the install 

Oh, there's an extra CMake step.  Sorry, I misread the original reply.

jonathonpenix (commented):
LGTM, thanks!

Once the Windows builders finish + we remove the `pull_request` bit I think this should be good to go if the build looks good.

efriedma-quic:
Nightly run looks good.  https://github.com/qualcomm/cpullvm-toolchain/actions/runs/22495247680?pr=202

jonathonpenix (approved):
Thanks Eli!

PR #213, opened 2026-03-02
Fix the Xqci baremetal configs

svs-quic:
This patch does the following things:

- Removes atomics from the march string and the variant name. 
- Adds DISABLE_THREADS: ON to disable threads.


pzhengqc:
The changes LGTM, just curious why we decide to disable threads?

qualcomm-software/embedded-multilib/json/variants/riscv32im_xqci_ilp32_nothreads_nopic.json
```diff
@@ -2,8 +2,8 @@
     "args": {
         "common": {
             "TARGET_ARCH": "riscv32",
-            "VARIANT": "riscv32ima_xqci_ilp32_nopic",
-            "COMPILE_FLAGS": "-march=rv32ima_zba_zbb_zbs_zca_zcb_zilsd_zclsd_xqcia_xqciac_xqcibi_xqcibm_xqcicli_xqcicm_xqcics_xqcicsr_xqciint_xqciio_xqcilb_xqcili_xqcilia_xqcilo_xqcilsm_xqcisim_xqcisls_xqcisync_xqccmp -mabi=ilp32 -fno-pic",
```
jonathonpenix, line 6: I don't mind this change since it aligns with other configs, but FWIW I'm not sure this is a no-op change. Even in the no-pic configs, compiler-rt (and maybe libc++?) are built under pic (mostly just in keeping with how things used to work). Not sure if this flag was overriding that or not.

Might be worth checking and mentioning in the commit message if this is a change in behavior (or doing as a separate change)

svs-quic:
> The changes LGTM, just curious why we decide to disable threads?

The initial request from @apazos was to not have threading.

qualcomm-software/embedded-multilib/json/variants/riscv32im_xqci_ilp32_nothreads_nopic.json
jonathonpenix: `-fPIC` from compiler-rt comes after `-fno-pic` from our compile flags it seems, so this should be a no-op change.

jonathonpenix (approved):
I'm happy with this if we want this as no threads.

Thanks!

apazos:
The linker is not optimizing PIC accesses when it is not needed for static builds, so we should not enable PIC everywhere, it might impact code size/performance. Did not hear about any request to support threads either.

jonathonpenix (approved):
LGTM, thanks Sudharsan!

FWIW too since it was re-added, I don't mind either way at this point on the fno-pic thing--the flag doesn't actually have any impact here IIUC (so it isn't really hurting anything, but doesn't necessarily help either).

We can clean it up if/when we either make compiler-rt "truly" -fno-pic or when we make everything PIC.

PR #216, opened 2026-03-02
[cpullvm][Github] Upload the toolchain in precheckins

jonathonpenix:
Retain it for only 24 hours to hopefully prevent issues with consuming too much storage.

Also upload it before testing--even if unit tests fail, it might be helpful to have access to the toolchain if it was built successfully.


apazos (commented):
Thanks, this should be temporary to allow validations before merging patches.

apazos (approved):

PR #219, opened 2026-03-02
[cpullvm][Github] Upload the toolchain in precheckins

jonathonpenix:
Retain it for only 24 hours to hopefully prevent issues with consuming too much storage.

Also upload it before testing--even if unit tests fail, it might be helpful to have access to the toolchain if it was built successfully.


jonathonpenix:
This is a cherry-pick from main, so I'm going to go ahead and push this through since we have people trying to do builds where this is helpful/was requested.

PR #224, opened 2026-03-03
Rel wf arg fix

navaneethshan:


navaneethshan:
--branch-repo is only valid after the subcommand release-workflow, so rearranged the argument.

jonathonpenix (approved):
Please include the "--branch-repo is only valid after the subcommand release-workflow, so rearranged the argument." bit as part of the message when merging

PR #228, opened 2026-03-03
[cpullvm] Build some aarch64 configs using initial exec TLS model

jonathonpenix:
local-exec is the default TLS model in picolibc when thread-local-storage is configured. However, we have a few cases where local-exec cannot be used, so change the configs used in those cases to be initial-exec.

This adds a new `TLS_MODEL` variable to control this.

There's a few things to note here:
1. This is intentionally done without multilib support for now as we don't want to be shipping separate per-TLS-model configs right now.
2. Ideally we'd use a consistent TLS model across most/all aarch64 configs (or all configs in general) but a) we're missing some TLS model optimizations in the linker (initial-exec to local-exec when the pre-requisites are met, etc.) and b) we've recently had some correctness issues pop up when testing initial-exec configs. So limit this to only the configs where we know we need this.
3. The TLS_MODEL option is only handled by picolibc. Other projects (compiler-rt, libc++) don't force this in a way that causes issues (or, generally don't seem to use TLS for the things we're building). So, leave those projects alone for now and address only picolibc. This should be easy to extend if we find out those projects need this set as well.


qualcomm-software/embedded-runtimes/CMakeLists.txt
```diff
@@ -494,6 +503,7 @@ if(C_LIBRARY STREQUAL picolibc)
             --buildtype=${LIBRARY_MESON_BUILD_TYPE}
             -Dsingle-thread=${single_thread}
             -Dthread-local-storage=${thread_local_storage}
+            -Dtls-model=${TLS_MODEL}
```
jonathonpenix, line 506: Just a side note: `tls-model` is and is documented in picolibc to be a no-op when `thread-local-storage` is not selected.

jonathonpenix:
Closing in favor of https://github.com/qualcomm/cpullvm-toolchain/pull/242

PR #231, opened 2026-03-04
Build libomp.a as part of Linux runtimes.

efriedma-quic:
riscv32 excluded because it's not supported upstream. libc++ path is explicitly specified because clang can't find it from the sysroot path. Otherwise this should be straightforward.

### DIFF
--- a/qualcomm-software/scripts/build_linux_runtimes.sh
+++ b/qualcomm-software/scripts/build_linux_runtimes.sh
@@ -343,6 +343,37 @@ for VARIANT in "${VARIANTS[@]}"; do
       -S "${LLVM_BASE_DIR}/runtimes"
   ninja -C "${LIBCXX_BUILD_DIR}" install
 
+  # Install openmp
+  if [[ ! "${VARIANT_ARCH}" =~ riscv32 ]]; then
+    echo "Installing openmp for ${VARIANT}"
+    OPENMP_BUILD_DIR="${VARIANT_BASE_BUILD_DIR}/openmp"
+    cmake -G Ninja \
+        -DCMAKE_INSTALL_PREFIX="${VARIANT_TMP_SYSROOT}" \
+        -DCMAKE_SYSROOT="${VARIANT_TMP_SYSROOT}" \
+        -DCMAKE_BUILD_TYPE="${CMAKE_OPT_LEVEL}" \
+        -DCMAKE_C_COMPILER="clang" \
+        -DCMAKE_CXX_COMPILER="clang++" \
+        -DCMAKE_SYSTEM_NAME="Linux" \
+        -DCMAKE_TRY_COMPILE_TARGET_TYPE="STATIC_LIBRARY" \
+        -DCMAKE_ASM_COMPILER_TARGET="${VARIANT_TARGET}" \
+        -DCMAKE_C_COMPILER_TARGET="${VARIANT_TARGET}" \
+        -DCMAKE_CXX_COMPILER_TARGET="${VARIANT_TARGET}" \
+        -DCMAKE_ASM_FLAGS="${LIB_BUILD_FLAGS}" \
+        -DCMAKE_C_FLAGS="${LIB_BUILD_FLAGS}" \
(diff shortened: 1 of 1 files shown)


qualcomm-software/scripts/build_linux_runtimes.sh
```diff
@@ -343,6 +343,37 @@ for VARIANT in "${VARIANTS[@]}"; do
...
+        -DCMAKE_C_COMPILER="clang" \
+        -DCMAKE_CXX_COMPILER="clang++" \
+        -DCMAKE_SYSTEM_NAME="Linux" \
+        -DCMAKE_TRY_COMPILE_TARGET_TYPE="STATIC_LIBRARY" \
+        -DCMAKE_ASM_COMPILER_TARGET="${VARIANT_TARGET}" \
+        -DCMAKE_C_COMPILER_TARGET="${VARIANT_TARGET}" \
+        -DCMAKE_CXX_COMPILER_TARGET="${VARIANT_TARGET}" \
+        -DCMAKE_ASM_FLAGS="${LIB_BUILD_FLAGS}" \
+        -DCMAKE_C_FLAGS="${LIB_BUILD_FLAGS}" \
+        -DCMAKE_CXX_FLAGS="${LIB_BUILD_FLAGS} -stdlib=libc++ -I${VARIANT_TMP_SYSROOT}/include/c++/v1/" \
```
jonathonpenix, line 363: I'm fine with this, but as a side comment I'm a bit surprised we need the explicit `-I`--we should have `--sysroot=${VARIANT_TMP_SYSROOT}` in `LIB_BUILD_FLAGS` already (and set CMAKE_SYSROOT) and I'd assume clang knows how to find these headers from that.

(I saw you mentioned it in the commit message too, but still somewhat surprised--not sure what I'm missing)

qualcomm-software/scripts/build_linux_runtimes.sh
```diff
@@ -343,6 +343,37 @@ for VARIANT in "${VARIANTS[@]}"; do
       -S "${LLVM_BASE_DIR}/runtimes"
   ninja -C "${LIBCXX_BUILD_DIR}" install
 
+  # Install openmp
+  if [[ ! "${VARIANT_ARCH}" =~ riscv32 ]]; then
+    echo "Installing openmp for ${VARIANT}"
+    OPENMP_BUILD_DIR="${VARIANT_BASE_BUILD_DIR}/openmp"
+    cmake -G Ninja \
+        -DCMAKE_INSTALL_PREFIX="${VARIANT_TMP_SYSROOT}" \
```
jonathonpenix, line 351: Does this get installed as something like `${VARIANT_TMP_SYSROOT}/lib/libomp.a`?

I don't have any opinion/objection, just curious since I think we have this tucked in an odd folder in our resource dir downstream.

(Related to this, I couldn't download the precheckin to check for some reason so I re-triggered the builds--sorry for the noise on that)

efriedma-quic: The install path is something like `aarch64-unknown-linux-musl/aarch64a/lib/libomp.a`, so it gets installed beside libc/libc++/etc.  The alternative is that we install to `lib/clang/22/lib/aarch64-unknown-linux-musl/aarch64a/libomp.a`, beside compiler-rt?  I don't really know what our rules are for choosing what to install where, but the current destination seems more conventional.

qualcomm-software/scripts/build_linux_runtimes.sh
efriedma-quic: It looks like the clang search code searches beside clang itself, then it searches `$SYSROOT/usr/include` and `$sysroot/usr/local/include`.  (Generic_GCC::addLibCxxIncludePaths.  Didn't verify that's the right code, but I think it is?)

qualcomm-software/scripts/build_linux_runtimes.sh
jonathonpenix: Yeah, next to libc/libc++ sounds good to me if the openmp build itself isn't opinionated in some way.

jonathonpenix (approved):

apazos:
It may be preferable to build the OpenMP library using glibc instead of MUSL. But currently, our Linux build primarily serves for testing purposes with MUSL. Need to confirm the users of OpenMP before integrating it.

PR #236, opened 2026-03-05
[cpullvm] Install an 'empty' multilib.yaml at the base of every embedded variant

jonathonpenix:
Currently, the clang's Baremetal toolchain object has a few hardcoded RISC-V multilib variants that are used if there is no multilib.yaml that clang can find.

Importantly, our library layouts do not match the layout the hardcoded multilibs expect, so when we *do* match one, options like `--sysroot` pointing to our variants break.

This creates issues for us in two scenarios:
1. While building out the sysroot for each variant, we use `--sysroot` to source the just-built headers and libraries.
2. We want to provide users the ability to use `--sysroot` to point to any of our library variants.

Installing a minimal multilib.yaml that matches any flags and points into the same folder as what multilib uses as the "base" directory (for Baremetal, this is the sysroot currently) seems like the easiest way to handle this without changes to upstream.

The other thing to note here is that we write this file out at configure time rather than build time. I waffled on this a bit, but it is a) similar to how we handle some other dependencies (meson's cross files, the "real" multilib.yaml) and b) is very simple to implement without worrying about our build dependencies. So, just live writing it out at configure time.

### DIFF
--- a/qualcomm-software/embedded-runtimes/empty-multilib.yaml
+++ b/qualcomm-software/embedded-runtimes/empty-multilib.yaml
@@ -1,0 +1,10 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This file is a minimal multilib configuration that should match any flags and
+# point into "base" directory used for multilib (currently, the sysroot).
+
+MultilibVersion: '1.0'
+Variants:
+- Dir: .
+  Flags: []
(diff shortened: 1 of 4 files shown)


qualcomm-software/embedded-runtimes/empty-multilib.yaml
jonathonpenix, line 1: I don't really love "empty-multilib.yaml", but I'm not sure what else to call it. I don't really want to use "dummy-multilib.yaml" or something like that in our repo.

apazos: @jonathonpenix a few name prefix options: open, alternate, placeholder, backup

jonathonpenix: Hm, might just leave it as "empty".

jonathonpenix (commented):
Added a test for this so we can hopefully see it working or breaking, rather than just relying on the build itself.

qualcomm-software/embedded-runtimes/empty-multilib.yaml
lenary: `null`? `single`?

lenary: `minimal`?

lenary (approved):
LGTM. We can always move this to a `build` step (rather than a `configure` step) later.

I'm not worried about the name really.

PR #240, opened 2026-03-05
[cpullvm] Have THIRD-PARTY-LICENSES.txt reflect the projects enabled

jonathonpenix:
Previously, we always included text about the picolibc/musl/musl-embedded
licenses in THIRD-PARTY-LICENSES.txt but selectively included the licenses
in `third-party-licenses`. So, if any of those projects were omitted for
a given build, there'd still be text pointing to the license in
`third-party-licenses/` but no actual license.

That's a bit confusing, so just omit the text from THIRD-PARTY-LICENSES.txt
as well if the projects aren't used in the build.

### DIFF
--- a/qualcomm-software/CMakeLists.txt
+++ b/qualcomm-software/CMakeLists.txt
@@ -793,7 +793,16 @@ install(
     COMPONENT llvm-toolchain-docs
 )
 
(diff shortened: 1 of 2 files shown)


apazos (approved):

jonathonpenix:
Licenses in the nightly build here look right to me: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/22738311382?pr=240

PR #241, opened 2026-03-05
Expose -ftls-model as a multilib flag for AArch64/ARM

pzhengqc:
This allows multilib selection based on tls model.

### DIFF
--- a/qualcomm-software/patches/llvm-project/0004-multilib-Expose-ftls-model-as-a-multilib-flag-for-AA.patch
+++ b/qualcomm-software/patches/llvm-project/0004-multilib-Expose-ftls-model-as-a-multilib-flag-for-AA.patch
@@ -1,0 +1,56 @@
+From 5f1ce69e1cbd688424a4b2efb93972f44a5bc8f5 Mon Sep 17 00:00:00 2001
+From: Pengxuan Zheng <pzheng@qti.qualcomm.com>
+Date: Thu, 5 Mar 2026 15:58:03 -0800
+Subject: [PATCH] [multilib] Expose -ftls-model as a multilib flag for
+ AArch64/ARM
+
+This allows multilib selection based on tls model.
+---
+ clang/lib/Driver/ToolChain.cpp                  | 8 ++++++++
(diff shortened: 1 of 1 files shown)


qualcomm-software/patches/llvm-project/0004-multilib-Expose-ftls-model-as-a-multilib-flag-for-AA.patch
```diff
@@ -0,0 +1,56 @@
...
+diff --git a/clang/lib/Driver/ToolChain.cpp b/clang/lib/Driver/ToolChain.cpp
+index 703a3f8819cb..d9796e820080 100644
+--- a/clang/lib/Driver/ToolChain.cpp
++++ b/clang/lib/Driver/ToolChain.cpp
+@@ -229,6 +229,10 @@ static void getAArch64MultilibFlags(const Driver &D,
+       break;
+     }
+   }
++
++  const Arg *TLSModel = Args.getLastArgNoClaim(options::OPT_ftlsmodel_EQ);
```
jonathonpenix, line 22: I'm not entirely sure, but I wonder if there needs to be something more than just looking at the command line args? In that I think the TLS model used can vary based on other options like pic, pie, etc. But I don't actually know if that information is available to us here.

(I also don't know if there is a "general" default either or if it varies per-variable.)

pzhengqc: It seems to me that the -ftls-model is for user to force certain tls model? I am not sure if LLVM does more than that though. We could try pushing this upstream and get some feedback. Another option is to use multilib custom flags and that should work for the purpose of selecting the variant based on tls model too. Any preferences?

apazos: For this release, let's push a simple patch to default selected aarch64 configs to TLS Init_exec model. 

And then you push the clang patch upstream to discuss with community, out of the critical path.

apazos:
@pzhengqc please close this PR, not needed  anymore.

PR #242, opened 2026-03-05
Update some aarch64 baremetal variants to use initial-exec TLS model

pzhengqc:
This patch borrowed the `TLS_MODEL` variable Jonathon implemented in https://github.com/qualcomm/cpullvm-toolchain/pull/228.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -1,8 +1,8 @@
 {
     "libs": [
         {
-            "variant": "aarch64a",
-            "json": "aarch64a.json",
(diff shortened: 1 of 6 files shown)


apazos (approved):

jonathonpenix (approved):

jonathonpenix (changes requested):
These file names need to be updated as well, that should be what is causing the failures

pzhengqc:
> These file names need to be updated as well, that should be what is causing the failures

Good catch! Just updated the PR to fix the file names.

jonathonpenix (approved):
LGTM, hopefully this goes through. Thanks!

PR #247, opened 2026-03-06
Add glibc-based openmp build.

efriedma-quic:
This involves downloading a gcc toolchain, and building against it.

While I was adding in downloads, I made some improvements to our download infrastructure: we now cache downloads, and use sha256sum to verify them.

### DIFF
--- a/.github/workflows/linux-premerge.yml
+++ b/.github/workflows/linux-premerge.yml
@@ -53,6 +53,14 @@ jobs:
           # old build products if the runner doesn't do it itself.
           clean: true
 
+      - name: Cache Linux runtime downloads
+        uses: actions/cache@v5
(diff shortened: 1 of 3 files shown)


jonathonpenix (approved):
Thanks!

Two misc. thoughts:

1. The caching is nice (and we need to do that for some other things like ccache I think)--I'm guessing we want to do this in our nightlies as well, but I don't mind if we do that separately.
2. Right now all our Linux libraries are off-by-default, is that still fine for this?

efriedma-quic:
> Right now all our Linux libraries are off-by-default, is that still fine for this?

They're enabled on the buildbot, which seems like the important thing?  I don't think the default for local builds matters.

PR #249, opened 2026-03-06
[cpullvm] Install libedit-dev for LLDB

jonathonpenix:
Install this optional dependency [1] for LLDB.

It looks like all the other ones are present in our AArch64 Github runners,
except Lua which we also disabled downstream anyway. (And we have Python
for lldb explicitly disabled right now.)

Apparently we also didn't run the install_dependencies.sh in our nightly. I
think that makes sense in that we're copying the runtimes so didn't need
meson. This should have been added with originally, but I missed it and
apparently we already had swig available.

[1] https://lldb.llvm.org/resources/build.html#optional-dependencies

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -69,7 +69,7 @@ jobs:
         include:
           - build_script: build_copy_runtimes.sh
             test_script: test.sh
-            install_script: ''
+            install_script: install_dependencies.sh
             target_os: Linux-AArch64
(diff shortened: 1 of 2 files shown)


jonathonpenix:
editline support is turned on in the AArch64 Linux builds with this: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/22773540353?pr=249

apazos (approved):

PR #258, opened 2026-03-11
[cpullvm][Github] Don't run precheckins based on modifying .md files

jonathonpenix:
There isn't really a point to running our precheckins when updating .md files like our README, docs, etc.

So add these files to the list of paths to ignore in our precheckin builds.


apazos:
Would it skip qcom-preflight-checks? Maybe that one should still be run?

jonathonpenix:
The qcom-preflight-checks will still run, it is just the toolchain build + test that is skipped (the .yaml file where this change is made)

apazos (approved):

PR #262, opened 2026-03-11
[cpullvm][Github] Update actions for Node deprecation

jonathonpenix:
Our builds are getting warnings as some of the actions that we're using are using deprecated versions of Node.js. See ex: [1]. So, update the relevant actions.

upload/download were left at v6/v7, respectively, as these releases should solve the Node.js issue, and the newer releases are *very* new. So let's wait a bit before updating to those.

While here, I'm also flipping over to using specific SHAs to be consistent with LLVM per [2].

[1] https://github.com/qualcomm/cpullvm-toolchain/actions/runs/22929283277
[2] https://llvm.org/docs/CIBestPractices.html

### DIFF
--- a/.github/workflows/linux-premerge.yml
+++ b/.github/workflows/linux-premerge.yml
@@ -46,7 +46,7 @@ jobs:
 
     steps:
       - name: Checkout source
-        uses: actions/checkout@v4
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           repository: ${{ github.event.pull_request.head.repo.full_name }}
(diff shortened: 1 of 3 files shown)


apazos (approved):
Do you plan to add this change to release 22.x? Need to check if we will have any builder failures then.

jonathonpenix:
I'd lean towards leaving this out of the branch at this point.

If we need to do a patch release >1 year down the line or something, we'd have to make this change but for now it isn't necessary.

jonathonpenix:
These test failures are also present in eld's latest Windows nightly, they're unrelated: https://github.com/qualcomm/eld/actions/runs/23034099797/job/66898183766

PR #263, opened 2026-03-11
[cpullvm] Specify exact Windows image we're using

jonathonpenix:
Align with guidance from [1] and what we do with the rest of our images in specifying the specific version we're using rather than 'latest'.

Hopefully this should prevent any surprises.

[1] https://llvm.org/docs/CIBestPractices.html


apazos (approved):

jonathonpenix:
FWIW, I'm planning on excluding this from the 22.x branch as well.

jonathonpenix:
Failures are present in the most recent eld Windows nightly as well, they should be unrelated: https://github.com/qualcomm/eld/actions/runs/23034099797/job/66898183766

PR #278, opened 2026-03-17
[cpullvm][Github] Add check for changes outside allowed files

jonathonpenix:
Generally, we're only allowed to make changes in a handful of locations
(.github, qualcomm-software, and a few documentation files). Add a new workflow
to enforce this.

Note that this is heavily based on ATfE's `check_downstream_changes.py` and we
use the same approach (and some of the same code!) in defining a pathspec file
that notes where we are allowed to make changes. This should be helpful once
we start syncing changes and it came up as being a useful thing when adding
our CHANGELOG in the past.

### DIFF
--- a/.github/workflows/check_modified_files.yml
+++ b/.github/workflows/check_modified_files.yml
@@ -1,0 +1,35 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries. 
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Check whether a PR makes changes only in files that we are allowed
+# to modify.
+
+name: Check Modified Files
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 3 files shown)


jonathonpenix:
Examples of the check:
- Changes that should be allowed: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/23222240168/job/67497204913?pr=279
- Changes that should be rejected: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/23222244399/job/67497218161?pr=280

(Although I'm going to slightly change the wording for when we reject files)

apazos (approved):

navaneethshan (approved):

jonathonpenix:
As a drive-by comment, I think there's things we can do that would make this a bit more friendly/fancy (like leaving a comment, etc.), but this seems like an ok stepping stone that doesn't require any work setting up Github permissions or whatnot.

PR #281, opened 2026-03-19
[cpullvm][Github] Add Zephyr Twister tests as a dispatchable workflow

jonathonpenix:
These tests take quite some time to run, so for now put them on Github's
runners to prevent tying up our own runners for hours a time. We can split
up these tests to run in parallel pretty easily, so we aren't penalized too
heavily for doing so. Right now each build takes ~2-2.75 hours. Also use
Zephyr's containers to avoid having to track specific dependencies.

Importantly too, this is tracking mainline Zephyr right now since we need
to manually patch eld support in. It's not clear to me what we'd want
longer term, but for now this seems like an ok starting point.

Also, leave this as only a workflow_dispatch-able workflow for now--more
ergonomic ways of launching this can come later.

### DIFF
--- a/.github/workflows/zephyr-twister-tests.yml
+++ b/.github/workflows/zephyr-twister-tests.yml
@@ -1,0 +1,146 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: Zephyr Twister Tests
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '.github/workflows/zephyr-twister-tests.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  build-Linux-x86-toolchain:
+    if: github.repository == 'qualcomm/cpullvm-toolchain'
+    name: Build toolchain
+    runs-on: self-hosted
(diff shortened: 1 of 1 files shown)


jonathonpenix:
Completed example of the tests: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/23360109497?pr=281

The per-subset build time is around ~2-2.75 hours, which seems like an ok place to be. Having this split up might be a bit annoying depending on what people are trying to do with the outputs, but I think we can merge the .xml files after the fact to make this easier. Will experiment with that.

apazos (approved):
Thanks Jonathon, this is very useful to track any Zephyr breakage with LLVM.
The only things that occurred to me:
- prioritize running kernel tests since these failures are blockers in upstream Zephyr- or maybe look for failing kernel tests in the merged log is fine, and it is what you have in this PR.
- build with a specific CPULLVM release - whoever launches the workflow can decide the CPULLVM branch to be used

jonathonpenix:
I think for now building more than just the kernel tests is ok--it's not too difficult to search for these if someone wants and we probably do want to look at non-kernel failures at some point too.

Might be worthwhile at some point to let people override this (or what platforms, Zephyr version, etc. is tested), but I think that can happen later as QOL improvements.

PR #283, opened 2026-03-25
[RISC-V] Add optimized memset/memcpy/strcpy/strcmp for Xqci

RamNalamothu:
The optimized implementations for Xqci will override the other existing corresponding varients when Xqci extenions are enabled.

The orverriding happens using the interface implemented in the upstream Picolibc pull requests 1090, 1092 and 1098.


RamNalamothu:
@jonathonpenix @apazos @svs-quic

This PR will need https://github.com/qualcomm/cpullvm-toolchain/pull/275 to apply cleanly and hence a standalone build applying only this PR will fail, which is expected.

I made this PR to contain only the changes explicit to XQCI, so that the other PR patch file can be deleted when we upgrade Picolibc and we don't need to modify this PR, but this will apply cleanly on the upgraded Picolibc, and it will be easier to manage by just having only two patch files in this manner.

PR #287, opened 2026-03-26
Updated the pre flight check yml file

sudeeshna21:
Running untrusted code on the pull_request_target trigger may lead to security vulnerabilities. These vulnerabilities include cache poisoning and granting unintended access to write privileges or secrets.

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

We should update all usage of pull_request_target in all workflow files and also update qualcomm-preflight-check to the latest.

### DIFF
--- a/.github/workflows/qcom-preflight-checks.yml
+++ b/.github/workflows/qcom-preflight-checks.yml
@@ -1,31 +1,25 @@
-name: Qualcomm Preflight Checks
+name: QC Preflight Checks
+
 on:
-  # FIXME: these branch targets might need to be revisited to make this work
-  # as we create new release branches.
-  pull_request_target:
-    branches:
-      - 'qualcomm-software'
-      - 'release/qualcomm-software/**'
+  pull_request:
+    branches: [qualcomm-software, release/qualcomm-software/**]
   push:
-    branches:
-      - 'qualcomm-software'
-      - 'release/qualcomm-software/**'
+    branches: [qualcomm-software, release/qualcomm-software/**]
   workflow_dispatch:
 
-permissions:
- contents: read
- security-events: write
-
 jobs:
-  qcom-preflight-checks:
-    if: github.repository == 'qualcomm/cpullvm-toolchain'
-    uses: qualcomm/qcom-reusable-workflows/.github/workflows/qcom-preflight-checks-reusable-workflow.yml@v1.1.4
+  preflight:
+    name: Run QC Preflight Checks
+    uses: qualcomm/qcom-reusable-workflows/.github/workflows/reusable-qcom-preflight-checks-orchestrator.yml@v2
     with:
-        # ✅ Preflight Checkers
-        repolinter: false                  # default: true
-        semgrep: true                      # default: true
-        copyright-license-detector: true   # default: true
-        pr-check-emails: true              # default: true
-        dependency-review: true            # default: true
-    secrets:
-      SEMGREP_APP_TOKEN: ${{ secrets.SEMGREP_APP_TOKEN }}
+      enable-semgrep-scan: true
+      enable-dependency-review: true
+      enable-repolinter-check: false
+      enable-copyright-license-check: true
+      enable-commit-email-check: true
+      enable-commit-msg-check: false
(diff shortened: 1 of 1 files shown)


mynameistechno (approved):
LGTM, please merge

.github/workflows/qcom-preflight-checks.yml
```diff
@@ -1,31 +1,25 @@
...
+    branches: [qualcomm-software, release/qualcomm-software/**]
   workflow_dispatch:
 
-permissions:
- contents: read
- security-events: write
-
 jobs:
-  qcom-preflight-checks:
-    if: github.repository == 'qualcomm/cpullvm-toolchain'
```
jonathonpenix, line 21: I don't think this should be removed--please either re-add this or explain why it isn't needed anymore (ex: if it is expected to work on forks out-of-the-box now, etc.)

jonathonpenix (dismissed):

.github/workflows/qcom-preflight-checks.yml
mynameistechno: It should work on forks - not sure why that was originally there

jonathonpenix: It was added as one of the checks (semgrep? something else? I forget) required additional security permissions that require extra setup (and generally getting notifications about failed runs in forks is just noise when it is coming out of the main repo).

mynameistechno: Maybe it was because the past workflow required an org level secret. We've removed that requirement

jonathonpenix: It is/was some Advanced Security/Dependency Graph setting: https://github.com/qualcomm/cpullvm-toolchain/pull/23

mynameistechno: Oh, yeah that is annoying. We added a check that skips the dep review if it's not enabled https://github.com/qualcomm/qcom-reusable-workflows/blob/main/.github/workflows/reusable-dependency-review.yml#L10

njjetha (approved):
LGTM!

navaneethshan (approved):

.github/workflows/qcom-preflight-checks.yml
navaneethshan: @jonathonpenix, this mean the qcom-preflight-checks will run on fork PRs too, but will not fail as it was happening before, hope this is okay?

jonathonpenix: That's good it now skips the dep review instead of outright failing.

FWIW, it isn't just PRs against a fork, it's also pushes--so any time you sync your fork to match the `qualcomm/cpullvm-toolchain` repo some of these fire (and bear in mind we pull 100+? commits per day from upstream where any failure in these checks is just noise). Granted, we haven't seen too many issues with this in the main repo (~5 in the past month or so?) so maybe it isn't a big deal, but I still see no upside doing this (both in general and as part of this change specifically).

But given we seem intent on turning these on in forks and https://github.com/qualcomm/cpullvm-toolchain/issues/394 is a thing, it's fine/we can always disable it again.

navaneethshan: @mynameistechno,  "push" can be removed? on pull-request is enough.

navaneethshan (approved):
Will resolve to unblock the PR uploads and work with IT on the foloow up work.

.github/workflows/qcom-preflight-checks.yml
navaneethshan: Resolved this conv to unblock the merge, will work with IT if it needs follow up work.

mynameistechno:
> Will resolve to unblock the PR uploads and work with IT on the foloow up work.

What does IT have to do with this?

navaneethshan:
> > Will resolve to unblock the PR uploads and work with IT on the foloow up work.
> 
> What does IT have to do with this?

Well, will work with your team to disable running this WF for upstream automerge commits which gets pushed directly without a PR.

PR #297, opened 2026-04-01
[patches] Remove Outdated Patches

lenary:
Patch | PR                       | Upstream     | Downstream
------|--------------------------|--------------|-------------
0002  | llvm/llvm-project#182350 | 8b9d47191468 | 0337deabbbb5
0003  | llvm/llvm-project#182220 | 605f4054bc42 | f0f9aa2619ea


lenary:
Do we want to merge this and fix the picolibc tests afterwards, or wait until they're fixed?

jonathonpenix:
Restarted the builder just to see what happens as Namish couldn't reproduce it locally.

Worst case, I don't mind if temporarily xfail this test or something to merge.

Edit: and I guess our builds are fundamentally broken anyway without this change, so if we merge this without xfailing that is fine with me too

jonathonpenix (approved):

navaneethshan (approved):

navaneethshan:
I suggest merging this to allow the picolibc fix to be verified on tip.

PR #298, opened 2026-04-03
[cpullvm][Github] Add qemu_riscv32_xip to tested Zephyr boards

jonathonpenix:


jonathonpenix:
This includes https://github.com/qualcomm/cpullvm-toolchain/pull/297 now just for testing, needs to be removed before merge

jonathonpenix:
Turns out this only results in...2 extra tests being run, but it does give a tiny bit of extra coverage for ALIGN_WITH_INPUT I think so it might be fine to keep for now.

Long term we should probably just do ex: `-a riscv32` anyway rather than specific platforms, so not too worried about the number of boards getting out of hand or something.

apazos:
What does xip board stand for in Zephyr?
And -a will try to run test for all boards? Shouldn't we just enable the qemu tests since we can't test all boards?

jonathonpenix:
> What does xip board stand for in Zephyr?

XIP is "execute in place", https://docs.zephyrproject.org/latest/kconfig.html#CONFIG_XIP has a bit more info too

> And -a will try to run test for all boards? Shouldn't we just enable the qemu tests since we can't test all boards?

Yeah, `-a` will enable all tests for the selected arch. I think there's ex: a non-QEMU simulator that some things can run on, and for some things there's tests for boards that are expected to build but won't run, etc.

In the short term (or for cpullvm purposes) I don't have an issue with sticking to just the qemu boards, but my assumption is at some point ex: the LLVM SDK should be clean for all these tests (the GCC SDK is expected to be clean). And, we should probably make sure eld is clean against these at some point if we're trying to get it into the SDK.

So at a certain point we might want to move beyond just qemu tests. But, qemu isn't clean, so that is probably best left for later.

jonathonpenix:
Putting this on top of https://github.com/qualcomm/cpullvm-toolchain/pull/304 for now to hopefully avoid the space errors.

jonathonpenix:
Edit: whoops, pasted this on the wrong ticket sorry, ignore this

apazos (approved):

jonathonpenix:
Passing build on top of #304: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/24115152695/job/70365462458?pr=298

Will rebase to remove #304 and merge

PR #302, opened 2026-04-07
[cpullvm] xfail failing picolibc math_handling tests

jonathonpenix:
This test started unexpectedly failing in our rv32 configs without 'f' when pulling in new LLVM commits.

It's been a bit and we're still investigating this, so xfail the test to unblock the builder.


apazos (approved):
I have created internal Jira to track the fix and re-enablement of the test.

PR #305, opened 2026-04-08
[cpullvm][Github] Prune Github workflows from upstream

jonathonpenix:
In bringing in our latest batch of upstream commits, we brought in more workflows from upstream. Clean these up again.

Hopefully this should be the last time before we can ignore these in automerge.


navaneethshan:
Yes, we will stop bringing .github files during automerge.

navaneethshan (approved):

PR #308, opened 2026-04-09
ci: add nightly schedule for workflow

navaneethshan:


jonathonpenix (approved):
Can we do this for the native runtimes nightly too?

Don't mind if it is same PR or a follow-up

navaneethshan:
> Can we do this for the native runtimes nightly too?
> 
> Don't mind if it is same PR or a follow-up

Done.

jonathonpenix (approved):
Thanks!

PR #309, opened 2026-04-10
[cpullvm] Add additional variants for testing Zephyr

jonathonpenix:
Two variants added:
* rv32gc ilp32d - Zephyr uses this extension/abi combo for quite
  a few hardfloat tests and we didn't have a compatible variant previously.
* armv7m nopic - Our existing variant is built with -fPIC and, for tests on
  qemu_cortex_m3 at least, Zephyr has linker script snippets that /DISCARD/
  the .got. eld currently can't relax referencing the .got, leading to
  errors when building. So for now provide a nopic variant for testing. Note
  also that, since this is for Zephyr only, we don't build this for
  musl-embedded. This config should also be removed once this
  optimization/relaxation is implemented in eld.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -56,6 +56,12 @@
             "flags": "--target=thumbv7m-unknown-none-eabi -mfpu=none"
         },
         {
+            "variant": "armv7m_soft_nofp_nopic",
+            "json": "armv7m_soft_nofp_nopic.json",
+            "flags": "--target=thumbv7m-unknown-none-eabi -mfpu=none -fno-pic",
+            "libraries_supported": "picolibc"
+        },
+        {
             "variant": "riscv32imc_ilp32_nothreads_nopic",
             "json": "riscv32imc_ilp32_nothreads_nopic.json",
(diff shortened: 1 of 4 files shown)


.github/workflows/zephyr-twister-tests.yml
```diff
@@ -5,9 +5,7 @@ name: Zephyr Twister Tests
 
 on:
   workflow_dispatch:
-  pull_request:
-    paths:
-      - '.github/workflows/zephyr-twister-tests.yml'
```
pzhengqc, line 10: What's this change for?

jonathonpenix: It's just a temporary change to trigger the Zephyr tests on this PR

pzhengqc (approved):
LGTM, thanks, @jonathonpenix!

jonathonpenix:
Zephyr test run that shows the changes working here: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/24266991288?pr=309

I don't see any of the cortex-m3 errors about discarding the .got, and similarly I don't see the same issues with not having a matching multilib for rv32 ilp32d

apazos (approved):

PR #317, opened 2026-04-16
[cpullvm] Set CC/CXX in the environment for Windows builds

jonathonpenix:
There's a few different aims here:
* Setting CMAKE_C_COMPILER, etc. directly can be a bit confusing when looking through our logs. That value is not propagated when we add multilib and the runtimes as external CMake projects so these projects report MSVC as CMAKE_C[XX]_COMPILER. We don't *use* MSVC in these subprojects (we manually pass in/use the just-built LLVM toolchain) so there shouldn't be any functional change here, but it should hopefully avoid confusion if/when people look through logs.

  So, I think it is best we avoid recommending this if we can.
* This also aligns our Windows builds closer to what we recommend for Linux.

With this change, I think we can also consolidate our build instructions a bit to make them (hopefully) more clear/concise.

### DIFF
--- a/qualcomm-software/docs/building.md
+++ b/qualcomm-software/docs/building.md
@@ -68,34 +68,30 @@ currently-enabled embedded variants.
 > [build scripts](/qualcomm-software/scripts) for examples on how our toolchains are built and packaged
 > on different hosts.
 
-### Linux
-The commands in the sections below assume you are in the `cpullvm-toolchain/qualcomm-software` directory.
+### Configuring and building the toolchain
 
-The toolchain can be built directly with CMake.
+The commands in the sections below assume you are in the `cpullvm-toolchain/qualcomm-software` directory.
 
 All dependencies are assumed to be present in your `PATH`.
 
+First, populate the `CC` and `CXX` environment variables to tell CMake which toolchain to use:
+
+Linux:
 ```
-# Alternatively, CMAKE_C_COMPILER/CMAKE_CXX_COMPILER may be set directly
 export CC=clang
 export CXX=clang++
+```
+Windows:
+```
+$env:CC = 'clang-cl'
+$env:CXX = 'clang-cl'
+```
+
+Then, the toolchain can be configured directly with CMake and built:
+```
 mkdir build
 cd build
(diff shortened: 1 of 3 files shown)


jonathonpenix:
Examples from our logs:

From one of our [nightly runs](https://github.com/qualcomm/cpullvm-toolchain/actions/runs/24491362793/job/71576856730):
```
[6053/6556] Performing configure step for 'multilib-picolibc'
-- The C compiler identification is MSVC 19.44.35225.0
-- The CXX compiler identification is MSVC 19.44.35225.0
```

From [building with this change](https://github.com/qualcomm/cpullvm-toolchain/actions/runs/24491797442/job/71578157330?pr=317):
```
[6053/6556] Performing configure step for 'multilib-picolibc'
-- The C compiler identification is Clang 19.1.5 with MSVC-like command-line
-- The CXX compiler identification is Clang 19.1.5 with MSVC-like command-line
```

Again this is really a cosmetic change, but it makes us a bit more consistent/hopefully preempts some confusion around this.

jonathonpenix:
Successful Windows builds (eld failures unrelated):
* Nightly: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/24491797450?pr=317
* Native runtime nightly: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/24491797442?pr=317

apazos (commented):
Is this an issue only on Windows? I think users might be more familiar to setting -DCMAKE_C_COMPILER, -DCMAKE_CXX_COMPILER, so should we keep them and add a note about Windows?

jonathonpenix:
It's not specific to Windows, it's just we already use `CC`/`CXX` on Linux in our build scripts, etc.

There isn't a particularly good reason why we weren't doing this on Windows before--it's just that when I was rushing through things I wasn't sure if CMake was consistent about this on Windows

I'm also not necessarily opposed to going through CMAKE_C_COMPILER (we can always pass this through to subprojects or something), but I think it's worth being consistent between whatever we recommend for Linux/Windows

apazos (approved):

PR #320, opened 2026-04-20
[cpullvm] Include our docs folder in the installation

jonathonpenix:
We include our README, etc. which reference these files. So, just include them in the installation as well in case they're helpful to refer to.

Note that this is what Arm's toolchain does/did--I think I originally omitted this since we had no 'docs/' folder at the time and forgot to re-include it once we created one.

### DIFF
--- a/qualcomm-software/CMakeLists.txt
+++ b/qualcomm-software/CMakeLists.txt
@@ -794,6 +794,12 @@ install(
     COMPONENT llvm-toolchain-docs
 )
 
+install(
+    DIRECTORY docs
+    DESTINATION .
+    COMPONENT llvm-toolchain-docs
+)
+
 if(LLVM_TOOLCHAIN_C_LIBRARY STREQUAL picolibc)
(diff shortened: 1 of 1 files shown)


jonathonpenix:
As a misc. comment, the links from the README -> docs folder should work if you "render" the README. There are some broken links to ex: CONTRIBUTING.md which we don't include (and which uses links that work for the Github repo but not really an install).

lenary (approved):
LGTM. I can see an argument for just including the user docs, but i'm fine with including everything for simplicity.

jonathonpenix:
Hm, yeah, thinking about it I guess the developer guide/build from source guide aren't that helpful in an installation. Maybe some of the dependency info can still be helpful?

I guess for now I'm just going to err towards including more than is strictly needed--if we have a ton of docs that aren't relevant in the future, we can revisit this I guess. It shouldn't be fundamentally more/less complicated to include only specific files.

PR #322, opened 2026-04-20
Added DockerFile to qualcomm-software.

pranav4330:
Please Review the DockerFile in /qualcomm-software/ci/docker path.  please refer to final commit: 57d25519e6797ceb9b8a7fe62ee615ab45ae8852


apazos:
this patch does more than just add the docker file. Can you separate the docker file review from the rest of the changes you made in this commit. If you do more than adding docker file you should describe the additional changes in the commit message.

.github/workflows/image_runner.yaml
```diff
@@ -0,0 +1,27 @@
+name: Pull cpullvm-toolchain image
```
apazos: This change is unrelated to adding the docker file, it can go in separate commit if we get permission to distribute the docker image + docker image script.

.github/workflows/image_runner.yaml
```diff
@@ -0,0 +1,27 @@
+name: Pull cpullvm-toolchain image
+
```
apazos: Similar to the comment above.

.gitkeep
apazos, line 1: similar comment as above.

qualcomm-software/ci/docker/README.md
apazos: Qualcomm LLVM (cpullvm) toolchain --> with the CPULLVM toolchain

apazos: ARM cross-compilers? How above RISC-V? It is not only about ARM. Remove this whole line...
Better phrasing: It includes the prerequisites outlined in Section  qualcomm-software/docs/building.md#prerequisites  for building an LLVM-based toolchain and the runtimes.

qualcomm-software/ci/docker/Dockerfile
apazos, line 30: The software required to build LLVM is described here:https://llvm.org/docs/GettingStarted.html#software
Why do we have more dependences?

apazos, line 37: These steps are not part of the build instructions for developers, why are they needed in the docker image?
I expect the script for building the docker image to be very close to how developers would build without the docker image as explained at qualcomm-software/docs/building.md and

qualcomm-software/ci/docker/README.md
apazos: This path is TBD, if we get approval.

apazos: the project name is CPULLVM Toolchain

apazos: Fix project name

apazos: This section is empty, need to be populated.

apazos: This section is empty, need to be populated.

apazos: Again, all the above session should basically include the dependences listed at
qualcomm-software/qualcomm-software/docs/building.md and https://llvm.org/docs/GettingStarted.html#software
and nothing else.
I do not understand why we have more dependencies listed here.

.github/workflows/toolchain_runner.yaml
```diff
@@ -0,0 +1,24 @@
...
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run LLVM toolchain build inside container
+        run: |
+          docker run --rm ghcr.io/pranav4330/cpullvmtoolchain22.x:ubuntu24.04 \
+            bash -c "ninja llvm-toolchain && ninja install llvm-toolchain"
```
apazos: Why are we not using the instructions qualcomm-software/qualcomm-software/docs/building.md#linux

jonathonpenix: I don't think we should have this at all--if we're going to build in the Docker container, it should be integrated in a consistent way with our other builds (ie just another matrix entry with the container specified)

qualcomm-software/ci/docker/Dockerfile
jonathonpenix: I'm also strongly in favor of removing this--I don't think we want to be messing with the system defaults when there isn't really a reason to.

jonathonpenix, line 32: Just eyeballing this, my guess is we're missing the optional LLDB dependencies as well

jonathonpenix, line 32: What about QEMU? xqci QEMU in particular is critical here

jonathonpenix (changes requested):
As a high level comment, if we're going the Docker route, I think our goal should be that the docker container serves *only* as a controlled environment to build in--there shouldn't be anything special about how we are building otherwise.

This is roughly in line with what Ana mentioned, but basically what I'd expect here is the dockerfile and if/when we decide to go this route/have the image accessible, there should be a `container:` + new matrix entry added in our usual build flows (with no special cmake or ninja commands, etc.) and the rest should be the same. (Or, a similar change should be added just for testing.)

Please do take a look at how upstream handles some of this in their testing:
* https://github.com/llvm/llvm-project/tree/main/.github/workflows/containers
* https://github.com/llvm/llvm-project/blob/e08c1c591fe8f41177afc98df5e3b61aadc66fe5/.github/workflows/ci-post-commit-analyzer.yml#L38-L39

A few misc. notes as well:
- Please note that commits must be signed off (see DCO failure, commit with `git commit -s ...` or add it manually) and please fix the address that you are using to commit with--it should be your qti email.
- Please update the PR title/description with a title/description that actually describes what the goal of the PR is, etc. https://llvm.org/docs/DeveloperPolicy.html#commit-messages is a good place to look for guidance. These commits also need to be squashed before or when merging--the PR title/description should reflect what that final commit message title/body should look like.
- In the future, please consider squashing WIP commits, etc. down into the logical commits you want to use before/when creating the PR or marking it ready for review (later commits addressing feedback, etc. can obviously be added and squashed later).
- Ana mentioned it, but please make sure all the .gitkeep files are removed--these shouldn't be needed.

I think there's a few more nits for later, but we can revisit those once the larger items are addressed

PR #323, opened 2026-04-22
[cpullvm] - Prepare for LIBCXX tests enablement ARM. 

pawan-nirpal-031:
Manage fixes and split testing script to run check-cxx separately to reduce the testing time for libcxx tests.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/variants/armv7a_hard_vfpv3.json
+++ b/qualcomm-software/embedded-multilib/json/variants/armv7a_hard_vfpv3.json
@@ -20,7 +20,7 @@
             "ENABLE_CXX_LIBS": "ON",
             "ENABLE_LIBC_TESTS": "ON",
             "ENABLE_COMPILER_RT_TESTS": "ON",
-            "ENABLE_LIBCXX_TESTS": "OFF"
+            "ENABLE_LIBCXX_TESTS": "ON"
         }
     }
 }
--- a/qualcomm-software/embedded-runtimes/CMakeLists.txt
+++ b/qualcomm-software/embedded-runtimes/CMakeLists.txt
@@ -235,12 +235,15 @@ if(ENABLE_LIBC_TESTS OR ENABLE_COMPILER_RT_TESTS OR ENABLE_LIBCXX_TESTS)
 endif()
 
 set(compile_arch_flags "--target=${target_triple} ${COMPILE_FLAGS} --sysroot ${TEMP_LIB_DIR}")
+if(NOT C_LIBRARY STREQUAL musl-embedded)
+    # Apply the same TLS model to test compilation as is used for library builds.
+    # Without this, -fPIC variants default to general-dynamic TLS which emits
+    # __tls_get_addr calls that are not available on bare-metal.
+    string(APPEND compile_arch_flags " -ftls-model=${TLS_MODEL}")
+endif()
 # Compiling the libraries benefits from some extra optimization
 # flags, and requires a sysroot.
 set(lib_compile_flags "${compile_arch_flags} -ffunction-sections -fdata-sections")
-if(NOT C_LIBRARY STREQUAL musl-embedded)
-    string(APPEND lib_compile_flags " -ftls-model=${TLS_MODEL}")
-endif()
 
 # Generic target names for the C library.
 # Declare these now, since compiler-rt requires the 'install' dependency.
--- a/qualcomm-software/embedded-runtimes/test-support/xfails.py
+++ b/qualcomm-software/embedded-runtimes/test-support/xfails.py
@@ -190,6 +190,153 @@ def main():
                 "insufficient memory. It requires at-least 10MB RAM."
         ),
         XFail(
+            name="flat-container-oom-armv7m",
+            testnames=[
+                "std/containers/container.adaptors/flat.map/flat.map.capacity/size.pass.cpp",
+                "std/containers/container.adaptors/flat.set/flat.set.capacity/size.pass.cpp",
+            ],
+            result=NewResult.XFAILED,
+            project="libcxx",
+            variants=[
+                "armv7m_soft_nofp",
+                "armv7m_soft_nofp_nopic",
+                "armv7m_hard_fpv5_d16_nopic",
+            ],
+            description="flat.map/size and flat.set/size insert 1,000,000 elements at "
+                        "runtime, exceeding the 8MB RAM on armv7m bare-metal QEMU targets. "
+                        "armv7a/armv8 have 16MB RAM and pass. The test aborts with exit 1.",
+        ),
+        XFail(
+            name="sort-heap-complexity-armv7m",
+            testnames=[
+                "std/algorithms/alg.sorting/alg.heap.operations/sort.heap/complexity.pass.cpp",
+                "std/algorithms/alg.sorting/alg.heap.operations/sort.heap/ranges_sort_heap.pass.cpp",
+            ],
+            result=NewResult.XFAILED,
+            project="libcxx",
+            variants=[
+                "armv7m_soft_nofp",
+                "armv7m_soft_nofp_nopic",
+                "armv7m_hard_fpv5_d16_nopic",
+            ],
+            description="sort.heap complexity tests use std::random_device which falls "
+                        "back to a fixed seed on armv7m bare-metal, causing the complexity "
+                        "assertion to fail. armv7a/armv8 pass. Exits with code 1.",
+        ),
+        XFail(
+            name="simd-unary-compiler-crash-armv7a",
+            testnames=[
+                "std/experimental/simd/simd.class/simd_unary.pass.cpp",
+            ],
+            result=NewResult.EXCLUDE,
+            project="libcxx",
+            variants=[
+                "armv7a_soft_neon",
+                "armv8_soft_neon",
+            ],
+            description="simd_unary.pass.cpp triggers a clang assertion failure "
+                        "(ScalarizeVecOp_VSETCC: expected v1i1 type) in "
+                        "LegalizeVectorTypes.cpp when compiling std::experimental::simd "
+                         "on ARMv7-A/ARMv8 with NEON. Compiler bug; exclude until fixed.",
+        ),
+        XFail(
+            name="long-double-picolibc-aarch64",
+            testnames=[
+                "std/strings/string.conversions/stold.pass.cpp",
+                "std/strings/string.conversions/to_string.pass.cpp",
+                "std/localization/locale.categories/category.monetary/locale.money.get/locale.money.get.members/get_long_double_overlong.pass.cpp",
+                "std/localization/locale.categories/category.numeric/locale.nm.put/facet.num.put.members/put_long_double.hex.pass.cpp",
+                "std/localization/locale.categories/category.numeric/locale.nm.put/facet.num.put.members/put_long_double.pass.cpp",
+                "std/input.output/iostream.format/output.streams/ostream.formatted/ostream.inserters.arithmetic/long_double.pass.cpp",
+            ],
+            result=NewResult.XFAILED,
+            project="libcxx",
+            variants=[
+                "aarch64a_tlsie",
+            ],
+            description="Long double is 128-bit on AArch64 and conversion between 128-bit "
+                        "types and strings is broken in picolibc. ARM 32-bit is unaffected "
+                        "(64-bit long double).",
+        ),
+        XFail(
+            name="aeabi-unwind-cpp-pr0-missing-arm",
+            testnames=[
+                "std/language.support/support.coroutines/end.to.end/",
+                "std/language.support/support.coroutines/coroutine.handle/coroutine.handle.noop/",
+                "std/language.support/support.coroutines/coroutine.handle/coroutine.handle.prom/",
+                "extensions/libcxx/odr_signature.exceptions.sh.cpp",
+            ],
+            result=NewResult.EXCLUDE,
+            project="libcxx",
+            variants=[
+                "armv8_soft_neon",
+                "armv7a_soft_neon",
+                "armv7a_soft_nofp",
+                "armv7a_hard_vfpv3",
+                "armv7m_soft_nofp",
+                "armv7m_soft_nofp_nopic",
+                "armv7m_hard_fpv5_d16_nopic",
+            ],
+            description="Tests that emit .ARM.exidx unwind tables referencing "
+                        "__aeabi_unwind_cpp_pr0, which is absent in the bare-metal "
+                        "semihosting runtime. Affects coroutine tests (which emit unwind "
+                        "tables even with -fno-exceptions) and odr_signature.exceptions "
+                        "(which mixes -fexceptions and -fno-exceptions TUs).",
+        ),
+        XFail(
+            name="at_exit",
+            testnames=[
+                "std/language.support/support.start.term/quick_exit.pass.cpp",
+            ],
+            result=NewResult.XFAILED,
+            project="libcxx",
+            description="at_quick_exit symbol is not found in the picolibc semihosting runtime.",
+        ),
+        XFail(
+            name="uchar-cuchar-xpass-picolibc",
+            testnames=[
+                "std/depr/depr.c.headers/uchar_h.compile.pass.cpp",
+                "std/strings/c.strings/cuchar.compile.pass.cpp",
+            ],
+            result=NewResult.PASSED,
+            project="libcxx",
+            description="uchar_h and cuchar compile tests are XFAIL upstream for picolibc "
+                        "(mbrtoc16 not defined), but picolibc does define mbrtoc16/c16rtomb "
+                        "so they pass on all our bare-metal variants. Suppress XPASS.",
+        ),
+        XFail(
+            name="cas-non-power-of-2-no-generic-atomic",
+            testnames=[
+                "std/atomics/atomics.types.generic/cas_non_power_of_2.pass.cpp",
+            ],
+            result=NewResult.EXCLUDE,
+            project="libcxx",
+            variants=[
+                "armv8_soft_neon",
+                "armv7a_soft_neon",
+                "armv7a_soft_nofp",
+                "armv7a_hard_vfpv3",
+                "armv7m_soft_nofp",
+                "armv7m_soft_nofp_nopic",
+                "armv7m_hard_fpv5_d16_nopic",
+            ],
+            description="cas_non_power_of_2.pass.cpp tests atomic operations on structs of "
+                        "sizes 3, 5, and 6 bytes, which require the generic (unsized) libcalls "
+                        "__atomic_load / __atomic_compare_exchange.  These lock-based fallbacks "
+                        "are not provided by libclang_rt.builtins.a on our bare-metal ARM targets.",
+        ),
+        XFail(
+            name="no-c8rtomb-verify-picolibc",
+            testnames=[
+                "std/strings/c.strings/no_c8rtomb_mbrtoc8.verify.cpp",
+            ],
+            result=NewResult.EXCLUDE,
+            project="libcxx",
+            description="Picolibc provides c8rtomb/mbrtoc8, so the verify test's expected "
+                        "errors ('no member named c8rtomb') do not fire. Exclude since the "
+                        "test premise does not apply to picolibc targets.",
+        ),
+        XFail(
             name="emulated crash signals",
             testnames=[
                 "aarch64/emupac.c",
--- a/qualcomm-software/scripts/test_libcxx.sh
+++ b/qualcomm-software/scripts/test_libcxx.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2025, Arm Limited and affiliates.
+# Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# Changes from Qualcomm Technologies, Inc. are provided under the following license:
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries. 
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# A bash script to run the libc++ tests for the Qualcomm embedded toolchain.
+#
+# These tests are kept separate from test.sh because ninja check-cxx takes
+# significantly longer to run than the rest of the test suite.  Run this
+# script in addition to test.sh when libc++ test coverage is required.
+#
+# The script assumes a successful build of the toolchain exists in the 'build'
+# directory inside the repository tree.
+
+set -ex
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( git -C "${SCRIPT_DIR}" rev-parse --show-toplevel )
+
+cd "${REPO_ROOT}"/build
+ninja  check-cxx-aarch64a_tlsie 
+ninja  check-cxx-armv7a_hard_vfpv3
+ninja  check-cxx-armv7a_soft_neon 
+ninja  check-cxx-armv7a_soft_nofp 
+ninja  check-cxx-armv7m_hard_fpv5_d16_nopic 
+ninja  check-cxx-armv7m_soft_nofp 
+ninja  check-cxx-armv7m_soft_nofp_nopic 
+ninja  check-cxx-armv8_soft_neon


pawan-nirpal-031:
@jonathonpenix Could you please review and add other reviewers as you see fit.

jonathonpenix:
At a glance, it looks like this is headed in the right direction, thanks Pawan!

Few misc. thoughts:

The libc++abi/libunwind tests are also included under the broader "enable libc++ tests". Right now they're running as we just run all enabled tests: https://github.com/qualcomm/cpullvm-toolchain/blob/8f6aeba8a85d6b9b5aff76c7a14fa14970e93021/qualcomm-software/scripts/test.sh#L20-L24

I'm fine if we enable this stuff incrementally, but I think we have a few options to do so. Things that make the most sense to me:
1. If we get down to just test failures, we can xfail and gradually re-enable tests and retain the `check-all*`
2. If we want to ex: merge this and address libcxxabi/libunwind build failures later, we can break up the `check-all-llvm-toolchain` into smaller components and exclude abi/unwind.

@pawan-nirpal-031 do you know/remember how long the libc++ tests were taking to run on your local machine?

Arm has these split out as they apparently take a long time to run, so we might want to do # 2 above even if we resolve build/test failures.

lenary:
I am not 100% sure how this ends up working, but are we correctly setting `long_tests` and `large_tests`? I think we should be if we're using qualcomm-software/embedded-runtimes/test-support/llvm-libc++-picolibc.cfg.in but I'm not sure if we are using that config or something else.

These prevent some of the really slow tests from running.

apazos:
They are already turned off in
embedded-runtimes/test-support/llvm-libc++-picolibc.cfg.inllvm-libc++-picolibc.cfg.in:
config.long_tests = False
config.large_tests = False

@pawan-nirpal-031, let's focus on fixing the issues first. We can merge the fixes without enabling the tests. Before activating the tests, we may need to split them into separate runs, as they seem to still take considerable time even with long/large tests disabled.

pawan-nirpal-031:
@apazos Yes I think that would be preferable. I've fixed the tests, some of them have to be unsupported due to memory limitations set to 2MB IIRC, those tests run into memory issues and I suppose aren't meant for embedded. Long double tests have to be xfailed through and through. But I will update this PR and we can discuss on the approach more. 

CC @jonathonpenix @lenary

pawan-nirpal-031:
```
Large test

These tests allocate very large amounts of memory (1,000,000 element containers, 1M-element vectors) that exceed what's available in the QEMU semihosting environment for bare-metal targets. The process aborts when malloc fails or the heap overflows.
Specifically:
- flat.map/size.pass.cpp: inserts 1,000,000 elements into a flat_map
- sort.heap/complexity.pass.cpp: allocates a 1M-element vector (1 << 20) and runs std::mt19937 over it
- eval_param.pass.cpp and simd_copy.pass.cpp: similarly large workloads
The bare-metal QEMU machine has a fixed, limited RAM (defined by the linker script picolibcpp.ld). When the test tries to allocate more than what's available, malloc returns null or the heap collides with the stack, causing an abort (signal 6 → exit 134).

std/algorithms/alg.sorting/alg.heap.operations/sort.heap/complexity.pass.cpp
 libc++-aarch64a_tlsie :: std/algorithms/alg.sorting/alg.heap.operations/sort.heap/ranges_sort_heap.pass.cpp
  libc++-aarch64a_tlsie :: std/containers/container.adaptors/flat.map/flat.map.capacity/size.pass.cpp
  libc++-aarch64a_tlsie :: std/containers/container.adaptors/flat.set/flat.set.capacity/size.pass.cpp
  libcxx/test/std/experimental/simd/simd.class/simd_copy.pass.cpp
````

```
UNSUPPORTED : Picolibc doesn't yet support stdio for long doubles for aarch64/arm/riscv

std/input.output/iostream.format/output.streams/ostream.formatted/ostream.inserters.arithmetic/long_double.pass.cpp
		  libc++-aarch64a_tlsie :: std/language.support/support.start.term/quick_exit.pass.cpp
		  libc++-aarch64a_tlsie :: std/localization/locale.categories/category.monetary/locale.money.get/locale.money.get.members/get_long_double_overlong.pass.cpp
		  libc++-aarch64a_tlsie :: std/localization/locale.categories/category.numeric/locale.nm.put/facet.num.put.members/put_long_double.hex.pass.cpp
		  libc++-aarch64a_tlsie :: std/localization/locale.categories/category.numeric/locale.nm.put/facet.num.put.members/put_long_double.pass.cpp
		std/strings/string.conversions/to_string.pass.cpp
		std/strings/string.conversions/stold.pass.cpp
                
```

```
Picolibc simply doesn't implement quick_exit or at_quick_exit yet. 
libcxx/test/std/language.support/support.start.term/quick_exit.pass.cpp
```

yet to get a hold on this one : libcxx/test/std/strings/c.strings/no_c8rtomb_mbrtoc8.verify.cpp

**Comment by apazos** [comment]
Hi @pawan-nirpal-031,

Besides tests failing, there are several ways to disable tests to control the time it takes to run the tests. Bringing them up here so we select the most appropriate one depending on the tests running time and the tests failing, so we can progressively enable them as we fix the issues:

1) Select the ninja check command to run in /qualcomm-software/qualcomm-software/scripts/test.sh
Right now, we enable ninja check-all-llvm-toolchain  (checks all LLVM/clang binaries and runtimes but excludes lib C++/abi/undwind  tests)

We can consider enabling tests for libc++ in parts, as done in ATfe, they skip ninja check-cxx which takes long, see ATfE /embedded/scripts/test.sh
ninja check-all  (checks all LLVM/clang binaries and runtimes including compiler-rt, lib C++/abi/undwind)
ninja check-all-llvm-toolchain (excludes lib C++ runtimes tests)
ninja check-cxxabi
ninja check-unwind
ninja check-cxx

2) ATfE has a separate /embedded/scripts/test_libcxx.sh where they only enable ninja check-cxx

3) This alternative allows to specify the particular project, variants and tests to disable
https://github.com/qualcomm/cpullvm-toolchain/blob/qualcomm-software/qualcomm-software/embedded-runtimes/test-support/xfails.py
We can check the tests disabled in ATfE /embedded/arm-runtimes/test_support/xfails.py

For example, if you only enable ninja check-cxxabi or ninja check-unwind or both, are there failures?

**Comment by pawan-nirpal-031** [comment]
@apazos Thanks for the detailed description, Yes I feel incrementally enabling would be the best option, ninja check-cxx are the ones that I am focusing on this PR. I imagine we don't want any source changes apart from changes in qualcomm-sotfware/* is that right? If yes I don't think there's much we can do to disable failing tests that need source changes, I'm wrestling with some riscv32* tests that are need support from upstream, Do you think we can push some of these  libcxx source changes upstream first? Or we can disable the tests now and then start upstreaming fixes into libcxx?

**Comment by apazos** [comment]
We have no custom LLVM changes in qualcomm-software and it keeps up with latest LLVM tip.
Which issue are you running into with riscv32* tests?
I am not sure the changes to tests are upstreamable (they reflect that in the particular runners/builders we are using they take a long time) and the way I see to disable the tests is along what ATfE does in /embedded/arm-runtimes/test_support/xfails.py.

**Comment by pawan-nirpal-031** [comment]
@apazos @jonathonpenix @lenary I have worked libcxx into a different test_libcxx.sh test script file similar to ATFE, All the tests locally pass, however I am omitting check unwind, abi and picolibc tests. The ETA for the libcxx tests locally is around 6-6:30 hours. The failure now seems to be purely CI timeout. ```The job has exceeded the maximum execution time of 6h0m0s```

**Comment by apazos** [comment]
@pawan-nirpal-031 this patch will have to be further cleaned up as some of these changes you are pushing in a separate PR. Also, might make sense to combine the xfailed test patch with the patch that enables the tests.

**Comment by pawan-nirpal-031** [comment]
> @pawan-nirpal-031 this patch will have to be further cleaned up as some of these changes you are pushing in a separate PR. Also, might make sense to combine the xfailed test patch with the patch that enables the tests.

Hi @apazos I'm sorry I didn't quite understand, you said we were only going to push the fixes and defer the enablement to a later PR. I will clean up the infra changes alright. Then as per me this PR should have xfails.py (fixes) and the separation of test_libcxx.sh. So if we also push the xfails to the enablement patch, this is just a CI change patch which separates the ninja check, is that what we intend to achieve? Apologies for the confusion.

**Comment by pawan-nirpal-031** [comment]
@apazos @jonathonpenix could you review it now. Also this seems to be failing the license check what would be the appropriate one?

File: qualcomm-software/embedded-runtimes/test-support/xfails.py [hunk 1]
```diff
@@ -164,36 +164,362 @@ def check_r52_warning():
...
             ],
             result=NewResult.XFAILED,
-            project="compiler-rt",
+            project="libcxx",
+            description="at_quick_exit symbol is not found in the picolibc semihosting runtime.",
+        ),
+        XFail(
+            name="long-double-picolibc-aarch64-riscv",
+            testnames=[
+                # These tests carry XFAIL: LIBCXX-PICOLIBC-FIXME && target=aarch64|riscv
```
jonathonpenix: I'm a bit confused on this--it sounds like they're xfail-ed for our targets upstream ("These tests carry XFAIL: LIBCXX-PICOLIBC-FIXME && target=aarch64|riscv...") and they are failing for us ("They do fail on our aarch64 and riscv bare-metal variants because...").

Why do we need to add the xfail here? Or am I misreading something?

qualcomm-software/embedded-runtimes/test-support/xfails.py
```diff
@@ -164,36 +164,362 @@ def check_r52_warning():
...
             ],
-            description="QEMU does not deliver crash signals. The emupac test uses "
-                "%expect_crash which requires the OS to deliver a signal back to the "
-                "test runner when PAC authentication fails. QEMU semihosting does not "
-                "support this capability. This test is also xfailed in the ATfE "
-                "toolchain (arm/arm-toolchain) for the same reason.",
+            description="picolibc does not yet support stdio for long doubles on aarch64 and "
+                        "riscv targets. These tests are XFAIL upstream for picolibc+aarch64/riscv "
+                        "and fail on our bare-metal variants for the same reason. "
+                        "ARM variants are handled separately by long-double-xpass-arm.",
```
jonathonpenix: Also confused along the same lines as above since `long-double-xpass-arm` is doing xpass, not xfail.

qualcomm-software/embedded-runtimes/test-support/xfails.py
```diff
@@ -164,36 +164,362 @@ def check_r52_warning():
...
+        ),
+        XFail(
+            name="to-string-long-double",
+            testnames=[
+                # to_string.pass.cpp calls test_float<long double>() which uses
+                # std::to_string(long double), internally calling snprintf with %Lf.
+                # picolibc does not support stdio for long doubles on aarch64/arm/riscv,
+                # so this subtest fails. The whole test is xfailed since we cannot
+                # disable a single subtest via lit without source changes.
+                "std/strings/string.conversions/to_string.pass.cpp",
```
jonathonpenix: This seems a bit odd to me as the description of `long-double-picolibc-aarch64-riscv` above seems to suggest that stdio for long doubles are implemented to some degree on Arm? Or are those tests skipped for 32bit Arm by some other means?

I also don't see this particular test XFAIL'ed in Arm's repo for any 32bit Arm targets (but do see it for AArch64, etc.)

qualcomm-software/embedded-runtimes/test-support/xfails.py
```diff
@@ -164,36 +164,362 @@ def check_r52_warning():
...
+            name="no-c8rtomb-mbrtoc8-picolibc",
+            testnames=[
+                "std/strings/c.strings/no_c8rtomb_mbrtoc8.verify.cpp",
+            ],
+            result=NewResult.EXCLUDE,
+            project="libcxx",
+            description="picolibc unconditionally declares c8rtomb() and mbrtoc8() in uchar.h, "
+                        "but _LIBCPP_HAS_C8RTOMB_MBRTOC8 is 0 (no source change to __config). "
+                        "The test expects these symbols to be absent, which is false for picolibc. "
+                        "Exclude rather than patching upstream libcxx headers.",
```
jonathonpenix: We already have a patch that I think should be doing exactly this: https://github.com/qualcomm/cpullvm-toolchain/blob/qualcomm-software/qualcomm-software/patches/llvm-project/0001-Define-_LIBCPP_HAS_C8RTOMB_MBRTOC8.patch

We inherited this from Arm's toolchain, and they don't seem to do anything special with this test.

So, seems like something is going wrong if this is needed (either this isn't needed, or our patch is broken somehow).

qualcomm-software/embedded-runtimes/test-support/xfails.py
```diff
@@ -164,36 +164,362 @@ def check_r52_warning():
...
+                "aarch64a_soft_nofp_tlsie",
+                "armv8_soft_neon",
+                "armv7a_soft_neon",
+                "armv7a_soft_nofp",
+                "armv7m_soft_nofp",
+                "armv7m_soft_nofp_nopic",
+            ],
+            description="stable_sort.pass.cpp runs exhaustive permutation tests up to N=8 "
+                        "and large-N sorts up to N=2053. Under QEMU emulation this exceeds "
+                        "the test timeout. ATfE runs on FVP which is faster; we exclude "
```
jonathonpenix: Is this test not running for RISC-V or is it actually passing?

Seems surprising to me that we wouldn't have to handle RISC-V variants similarly. (I assume RISC-V QEMU isn't that much faster than Arm/AArch64?)

qualcomm-software/embedded-runtimes/test-support/xfails.py
```diff
@@ -164,36 +164,362 @@ def check_r52_warning():
...
+        ),
+        XFail(
+            name="cmath-riscv-no-f",
+            testnames=[
+                "std/numerics/c.math/cmath.pass.cpp",
+            ],
+            result=NewResult.EXCLUDE,
+            project="libcxx",
+            variants=[
+                # cmath.pass.cpp exercises long double and floating-point math functions
```
jonathonpenix: Might be worth merging this with the AArch64 soft float one above since it sounds like the same issue

qualcomm-software/scripts/test.sh
```diff
@@ -21,4 +21,5 @@ REPO_ROOT=$( git -C "${SCRIPT_DIR}" rev-parse --show-toplevel )
 # but while we have a limited number of variants (and no compiler-rt
 # or libc++ testing enabled) we can run everything.
 cd "${REPO_ROOT}"/build
-ninja check-all-llvm-toolchain
+ninja check-all
```
jonathonpenix, line 24: Do you have a rough idea how long the libc++ tests for an individual variant take to run? Can you check on this?

Needs input from others, but it might be worthwhile including at least a couple "core" variants to be tested during premerge (and normal nightly?) checks and we can run the full suite (like what you did for test_libcxx.sh) separately. See ex: https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/scripts/test_pre_merge.sh

qualcomm-software/scripts/test_libcxx.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# If a test fails, lit will ordinarily return a non-zero result,
+# which prevents further testing. Setting the --ignore-fail option
+# will cause testing to continue, so that CI systems can get a
+# full set of results.
+# The lit test suites do not generate xml results by default.
+# This can be enabled with the --xunit-xml-output option. The file
+# written will be relative to the individual suite's build directly,
+# so the same name can be used for all files for consistency.
+export LIT_OPTS="--ignore-fail --xunit-xml-output=lit_results.junit.xml"
```
jonathonpenix: We might want to omit this part for now.

I think eventually we want something like this (we've discussed re-adding this into test.sh a few times as we've already been bitten a few times by the "exit at first fail" hiding other issues) but we'd need people to be aware to look at this (or, do some more handling of it to make the failures more obvious).

@apazos do you have thoughts/preferences here?

(I guess this is also a bit moot though until we actually schedule something to run these tests)

jonathonpenix (commented):
This looks pretty close to me, thanks a lot @pawan-nirpal-031 for working on this.

> Also this seems to be failing the license check what would be the appropriate one?

Apache 2 + LLVM is right, AFAIK that check assumes BSD 3 clause by default which is where our errors generally come from--it's something we have to sort out with the folks who maintain that. So, there's nothing you need to change here AFAIK

qualcomm-software/embedded-runtimes/test-support/xfails.py
jonathonpenix, line 193: I'm assuming something went wrong with a rebase?

Presumably the compiler-rt xfails shouldn't be deleted as part of this change

qualcomm-software/embedded-runtimes/test-support/xfails.py
jonathonpenix: @pawan-nirpal-031 it seems these xfails were removed entirely in the latest patch? Does that mean all targets are passing without these xfails?

qualcomm-software/scripts/test.sh
jonathonpenix: Ping

jonathonpenix (commented):
A few misc. comments:
1. There's a few updates that seem somewhat surprising--ex: `long-double-xpass-arm` used to be xpassing some tests for 32bit Arm variants but has been removed, same with ex: `stable_sort.pass.cpp`, etc. Are you seeing large swings in what is passing/failing? It makes reviews tricky, especially when there isn't really mention about what is changing or why.
2. With these changes, are you able to get a clean run of the libcxx tests? Something like what Garvit did here would be helpful to see for the libcxx tests (or a sample of the variants depending on how long things take to run): https://github.com/qualcomm/cpullvm-toolchain/pull/396
3. I think a rebase should fix the errors? Our nightlies are clean at least.

pawan-nirpal-031:
> @pawan-nirpal-031 it seems these xfails were removed entirely in the latest patch? Does that mean all targets are passing without these xfails?

@jonathonpenix these tests take a long time to run, I wanted to xfail them to investigate later, they take long enough that they don't report a pass or a fail. But I'm skeptical if CI will actually allow them to run that long. So I've removed them for now.

qualcomm-software/embedded-runtimes/CMakeLists.txt
```diff
@@ -235,12 +235,15 @@ if(ENABLE_LIBC_TESTS OR ENABLE_COMPILER_RT_TESTS OR ENABLE_LIBCXX_TESTS)
 endif()
 
 set(compile_arch_flags "--target=${target_triple} ${COMPILE_FLAGS} --sysroot ${TEMP_LIB_DIR}")
+if(NOT C_LIBRARY STREQUAL musl-embedded)
+    # On bare-metal static executables the local-exec TLS model is always
+    # correct. Without this, -fPIC variants use general-dynamic TLS which
+    # emits __tls_get_addr calls that are not available on bare-metal.
+    string(APPEND compile_arch_flags " -ftls-model=local-exec")
```
jonathonpenix: This isn't right--there are variants that are specifically/intentionally built under initial-exec to support non-static executables. We can't just not do that anymore.

If you're seeing an issue, please provide more information.

The `__tls_get_addr` issue described in the comment should already have been addressed (that's what the lines that were deleted are there for).

quic-garvgupt:
There are 7 variants for which libc testing is enabled but not libcxx tests.
- armv7a_hard_vfpv3
- armv7m_hard_fpv5_d16_nopic
- riscv32imc_ilp32_nothreads_nopic
- riscv32imc_zba_zbb_zbc_zbs_ilp32_nothreads_nopic
- riscv32im_xqci_ilp32_nothreads_nopic
- riscv32im_zba_zbb_zbc_zbs_zca_zcb_zcmp_ilp32_nothreads_nopic
- riscv64imc_lp64_nothreads_nopic

Pls enable libcxx tests for the above variants as well.

quic-garvgupt:
> There are 7 variants for which libc testing is enabled but not libcxx tests.
> 
> * armv7a_hard_vfpv3
> * armv7m_hard_fpv5_d16_nopic
> * riscv32imc_ilp32_nothreads_nopic
> * riscv32imc_zba_zbb_zbc_zbs_ilp32_nothreads_nopic
> * riscv32im_xqci_ilp32_nothreads_nopic
> * riscv32im_zba_zbb_zbc_zbs_zca_zcb_zcmp_ilp32_nothreads_nopic
> * riscv64imc_lp64_nothreads_nopic
> 
> Pls enable libcxx tests for the above variants as well.

FYI: These variants also has `ENABLE_CXX_LIBS` set to ON.

pawan-nirpal-031:
> > There are 7 variants for which libc testing is enabled but not libcxx tests.
> > 
> > * armv7a_hard_vfpv3
> > * armv7m_hard_fpv5_d16_nopic
> > * riscv32imc_ilp32_nothreads_nopic
> > * riscv32imc_zba_zbb_zbc_zbs_ilp32_nothreads_nopic
> > * riscv32im_xqci_ilp32_nothreads_nopic
> > * riscv32im_zba_zbb_zbc_zbs_zca_zcb_zcmp_ilp32_nothreads_nopic
> > * riscv64imc_lp64_nothreads_nopic
> > 
> > Pls enable libcxx tests for the above variants as well.
> 
> FYI: These variants also has `ENABLE_CXX_LIBS` set to ON.

Noted, added the two armv7s, will add riscv in the following patches due to CI timeout all versions cannot be accommodated.

quic-garvgupt:
Pls update the commit header to reflect the latest changes that this PR only enables test for ARM

qualcomm-software/embedded-runtimes/test-support/xfails.py
```diff
@@ -151,6 +151,233 @@ def check_r52_warning():
             ],
             description="Disable the tests for now while the issue is being fixed upstream (https://github.com/picolibc/picolibc/pull/1072).",
         ),
+        XFail(
+            name="flat-container-oom-armv7m",
+            testnames=[
+                # flat.map/size and flat.set/size insert 1,000,000 elements at runtime,
```
jonathonpenix: Please remove these comments that are just duplicates (or very near duplicates) of the description. Having this info/having to maintain this info twice isn't particularly useful.

qualcomm-software/embedded-runtimes/test-support/xfails.py
```diff
@@ -151,6 +151,233 @@ def check_r52_warning():
...
+                "armv7m_soft_nofp_nopic",
+                "armv7m_hard_fpv5_d16_nopic",
+            ],
+            description="Coroutine tests that instantiate real coroutine frames emit .ARM.exidx "
+                        "unwind tables referencing __aeabi_unwind_cpp_pr0, which is absent in "
+                        "the bare-metal semihosting runtime. Exclude for ARM variants built "
+                        "with ENABLE_EXCEPTIONS=OFF.",
+        ),
+        XFail(
+            name="odr-signature-exceptions-arm-nofp",
```
jonathonpenix: Seems like this should just be merged with coroutines-aeabi-unwind-arm if it is the same issue in terms of 32bit Arm runtimes built with `-fno-exceptions` not having `__aeabi_unwind_cpp_pr0`.

qualcomm-software/embedded-multilib/json/variants/armv7a_hard_vfpv3.json
```diff
@@ -20,7 +20,7 @@
             "ENABLE_CXX_LIBS": "ON",
             "ENABLE_LIBC_TESTS": "ON",
             "ENABLE_COMPILER_RT_TESTS": "ON",
-            "ENABLE_LIBCXX_TESTS": "OFF"
+            "ENABLE_LIBCXX_TESTS": "ON"
```
jonathonpenix, line 23: Just a note that this isn't included in #341 as it is a "new" config.

Since this is already here I'm fine if it stays **if** no other changes are needed, but this should be the only case where we're turning on new libc++ tests in .json.

And, if there are failures due to libc++abi needing xfails for this variant or something, please remove this and we'll add it in a follow-up.

qualcomm-software/scripts/test_libcxx.sh
```diff
@@ -0,0 +1,28 @@
...
+# directory inside the repository tree.
+
+set -ex
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( git -C "${SCRIPT_DIR}" rev-parse --show-toplevel )
+
+# Run libc++ tests using Ninja.
+cd "${REPO_ROOT}"/build
+ninja check-cxx
```
jonathonpenix: If the goal is to incrementally enable libc++ testing for only Arm/AArch64, this should be split this up rather than rely on it being disabled in the variant json files.

qualcomm-software/scripts/test.sh
jonathonpenix: To answer this, it seems to be about ~13min for each of the variants enabled currently.

Leaving any changes to the premerge/normal test script to be done separately.

jonathonpenix:
Also, I have no problem if you want to go ahead with doing Arm/AArch64 and RISC-V separately.

Once a run of these libc++ tests is clean, please post a link to the successful build in the PR here and remove the changes to test.sh.

Then hopefully we can get this merged at least and move onto RISC-V separately.

pawan-nirpal-031:
> Also, I have no problem if you want to go ahead with doing Arm/AArch64 and RISC-V separately.
> 
> Once a run of these libc++ tests is clean, please post a link to the successful build in the PR here and remove the changes to test.sh.
> 
> Then hopefully we can get this merged at least and move onto RISC-V separately.

https://github.com/qualcomm/cpullvm-toolchain/actions/runs/28502704462/job/84483587649?pr=323

Build passes. I'll remove all the changes to test.sh also removing libcxx_test.sh.

pawan-nirpal-031:
Some Eld realted failure. Unrealted to the patch. 

********************
1 warning(s) in tests
Failed Tests (1):
  eld :: Common/standalone/VersionScript/VersionScript.test

qualcomm-software/embedded-runtimes/test-support/xfails.py
```diff
@@ -189,6 +189,169 @@ def check_r52_warning():
...
+                "std/localization/locale.categories/category.monetary/locale.money.get/locale.money.get.members/get_long_double_overlong.pass.cpp",
+                "std/localization/locale.categories/category.numeric/locale.nm.put/facet.num.put.members/put_long_double.hex.pass.cpp",
+                "std/localization/locale.categories/category.numeric/locale.nm.put/facet.num.put.members/put_long_double.pass.cpp",
+                "std/input.output/iostream.format/output.streams/ostream.formatted/ostream.inserters.arithmetic/long_double.pass.cpp",
+            ],
+            result=NewResult.XFAILED,
+            project="libcxx",
+            variants=[
+                "aarch64a_tlsie",
+                "aarch64a_soft_nofp_tlsie",
```
jonathonpenix: Please remove aarch64a_soft_nofp_tlsie, we don't build these C++ libraries.

qualcomm-software/embedded-runtimes/test-support/xfails.py
```diff
@@ -189,6 +189,169 @@ def check_r52_warning():
...
+                "std/depr/depr.c.headers/uchar_h.compile.pass.cpp",
+                "std/strings/c.strings/cuchar.compile.pass.cpp",
+            ],
+            result=NewResult.PASSED,
+            project="libcxx",
+            description="uchar_h and cuchar compile tests are XFAIL upstream for picolibc "
+                        "(mbrtoc16 not defined), but picolibc does define mbrtoc16/c16rtomb "
+                        "so they pass on all our bare-metal variants. Suppress XPASS.",
+        ),
+        XFail(
```
jonathonpenix: Please remove this, we don't build c++ libraries for aarch64a_soft_nofp_tlsie

jonathonpenix (commented):
> > Also, I have no problem if you want to go ahead with doing Arm/AArch64 and RISC-V separately.
> > Once a run of these libc++ tests is clean, please post a link to the successful build in the PR here and remove the changes to test.sh.
> > Then hopefully we can get this merged at least and move onto RISC-V separately.
> 
> https://github.com/qualcomm/cpullvm-toolchain/actions/runs/28502704462/job/84483587649?pr=323
> 
> Build passes. I'll remove all the changes to test.sh also removing libcxx_test.sh.

Thanks for posting the link.

Please re-add test_libcxx.sh. It should have the `ninja check-libcxx-<variant>` lines for the variants expected to be runnable after this patch. I know it isn't used anywhere currently, but we'll add a builder that uses this elsewhere.

jonathonpenix (approved):
LGTM, thanks!

apazos (approved):

apazos (approved):

jonathonpenix:
Just mentioning that a precheckin build was done with Linux libraries to work around unrelated issues before merging here and is/was clean: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/28613856967/job/84852677356

PR #328, opened 2026-04-27
[cpullvm] Un-xfail picolibc math_handling tests

svs-quic:
These were xfailed in qualcomm/cpullvm-toolchain#302 but no longer fail since the musttail patches causing the failures were reverted in the [llvm project](https://github.com/llvm/llvm-project/pull/191508).


lenary (approved):
LGTM

PR #330, opened 2026-04-28
Fix developer documentation to build the correct aarch64 variant

quic-garvgupt:


qualcomm-software/docs/developing.md
```diff
@@ -102,9 +102,9 @@ Then build as usual.
 When building the toolchain, specific subsets of embedded library variants to build can be selected by
 setting `LLVM_TOOLCHAIN_LIBRARY_VARIANTS`.
 
-For example, the command below would build only the 'aarch64a' and 'riscv32imac_ilp32' variants:
+For example, the command below would build only the 'aarch64a_tlsie' and 'riscv32imac_ilp32' variants:
 ```
-cmake .. -GNinja -DLLVM_TOOLCHAIN_LIBRARY_VARIANTS="aarch64a;riscv32imac_ilp32"
+cmake .. -GNinja -DLLVM_TOOLCHAIN_LIBRARY_VARIANTS="aarch64a_tlsie;riscv32imac_ilp32"
 ninja llvm-toolchain
```
apazos, line 109: Renaming is unnecessary and could hurt readability.
The intent of this text is to show how to restrict the variants being built.
The current variant names can remain, but we could clarify that each variant is built using configuration settings from its respective JSON file (that is where the TLS config is done), but I think we already mention it somewhere.

quic-garvgupt: I am not sure if specifying correct variant is compromising on the readability. We can still mention that names for the variant is coming from multilib.json file.   Additionally, specifying the commands that doesnt work out of the box doesn't give much confidence on the text and leave developers scrambling around to find the correct command. 

IMO the latest changes in codebase should also get reflected in the documentation as well.

lenary: I think you're both right:

- The "_tlsie" suffix is not obvious as to what it means.
- The commands should work.
- There should be a reference to the the fact these names come from the json names.

@quic-garvgupt the specific variants don't matter here, can we find one with a more obvious name? Maybe `aarch64a_pacret`? IMO pacret is a more well known concept for aarch64 than the TLS model names.

quic-garvgupt: Done

apazos: Thanks @quic-garvgupt!

apazos (approved):

lenary (approved):
LGTM

PR #337, opened 2026-05-05
[patch_repo] allow script to detect if patches have been applied already

bojle:
In the current implementation, re-running builds fails as git apply fails to apply patches to _deps/* due to the patches already being applied. This patch fixes that by detecting if patches have been applied using git apply --reverse checks.

### DIFF
--- a/qualcomm-software/cmake/patch_repo.py
+++ b/qualcomm-software/cmake/patch_repo.py
@@ -25,7 +25,7 @@ def main():
         "--method",
         choices=["am", "apply"],
         default="apply",
-        help="Git command to use. git am will add each patch as a commit, whereas git apply will leave patched changes staged.",
+        help="Git command to use. git am will add each patch as a commit, whereas git apply will leave patched changes staged and skip patches that are already applied.",
     )
     parser.add_argument(
         "--reset",
@@ -114,7 +114,9 @@ def main():
                 if args.three_way:
                     apply_check_args.append("--3way")
                 apply_check_args.append(str(current_patch))
-                p_check = subprocess.run(apply_check_args)
+                p_check = subprocess.run(
+                    apply_check_args, capture_output=True, text=True
+                )
 
                 if p_check.returncode == 0:
                     # Patch will apply.
@@ -129,7 +131,23 @@ def main():
                     p = subprocess.run(apply_args, check=True)
                     applied_patches.append(current_patch)
                 else:
-                    # Patch won't apply.
+                    # Patch does not apply cleanly. Check if it is already applied
+                    # by attempting a reverse check.
+                    reverse_check_args = git_cmd + [
+                        "apply",
+                        "--ignore-whitespace",
+                        "--reverse",
+                        "--check",
+                        str(current_patch),
+                    ]
+                    p_rev = subprocess.run(
+                        reverse_check_args, capture_output=True, text=True
+                    )
+                    if p_rev.returncode == 0:
+                        print(f"Skipping {current_patch.name} (already applied).")
+                        continue
+
+                    # Patch is not already applied and cannot be applied cleanly.
                     print(f"Unable to apply {current_patch.name}")
                     if args.restore_on_fail:
                         # Remove any patches that have already been applied.
--- a/qualcomm-software/docs/building.md
+++ b/qualcomm-software/docs/building.md
@@ -50,7 +50,7 @@ Generally, these patches are automatically applied, except for llvm-project whic
 patched manually. The below command assumes you are in the `cpullvm-toolchain` directory:
 
 ```
-python3 qualcomm-software/cmake/patch_repo.py --method apply qualcomm-software/patches/llvm-project
+python3 qualcomm-software/cmake/patch_repo.py qualcomm-software/patches/llvm-project
 ```
 
 Other projects (eld, picolibc, etc.) are checked out and patched automatically. If you prefer, you can check
--- a/qualcomm-software/docs/building.md
+++ b/qualcomm-software/docs/building.md
@@ -50,7 +50,7 @@ Generally, these patches are automatically applied, except for llvm-project whic
 patched manually. The below command assumes you are in the `cpullvm-toolchain` directory:
 
 ```
-python3 qualcomm-software/cmake/patch_repo.py qualcomm-software/patches/llvm-project
+python3 qualcomm-software/cmake/patch_repo.py --method apply qualcomm-software/patches/llvm-project
 ```
 
 Other projects (eld, picolibc, etc.) are checked out and patched automatically. If you prefer, you can check
--- a/qualcomm-software/docs/developing.md
+++ b/qualcomm-software/docs/developing.md
@@ -75,13 +75,12 @@ and libc++/libc++abi/libunwind libraries for the variant.
 ### Re-running CMake
 Unless you have manually checked out and patched CPULLVM's various dependencies (see "Manually checking out
 and patching dependencies" below), CPULLVM's CMake will automatically checkout and patch the dependencies'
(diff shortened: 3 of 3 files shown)


bojle:
I've implemented the detection logic for 'git apply' for now. I can implement something similar for 'git am' too.

qualcomm-software/cmake/patch_repo.py
navaneethshan, line 150: Why is this needed? GitHub workflow is going to start on a clean workspace everyime.

apazos:
Hi @bojle,

It seems you are trying to apply the patch files over and over in your local build, not following the instructions to skip patching: 
 
https://github.com/qualcomm/cpullvm-toolchain/blob/release/qualcomm-software/22.x/qualcomm-software/docs/developing.md#manually-checking-out-and-patching-dependencies 
(Set DFETCHCONTENT_FULLY_DISCONNECTED=ON)
 
https://github.com/qualcomm/cpullvm-toolchain/blob/qualcomm-software/qualcomm-software/docs/building.md#patching
(Do not repeat the command python3 qualcomm-software/cmake/patch_repo.py --method apply qualcomm-software/patches/llvm-project)
 
No need to change the github workflow, as Navaneeth pointed out, the workflow creates a new space every time

qualcomm-software/cmake/patch_repo.py
bojle: This is an issue while working locally. Consider this workflow:

We start with this:

```
cmake -DCMAKE_BUILD_TYPE=Release
ninja
```
Now, we feel the need to change the build type to debug, this is how we'd do it:

```
cmake -DCMAKE_BUILD_TYPE=Debug
ninja
```

the second invocation of cmake fails as the script tries to re-apply the latch, unless we explicitly specify the variable to switch off patch application. 

this PR makes the patch detection logic automatic

bojle:
Hi @apazos

Please see my other comment https://github.com/qualcomm/cpullvm-toolchain/pull/337#discussion_r3190940681

In local workflows, any change to cmake variables will trigger the patch application script which ultimately fails if its not the first application. Only workaround so far is the cmake variable to disable it (which is a manual process), if its just a python script applying the patch, why can't it check and fail gracefully when its not the first application?

apazos:
The proper cmake command should include -DFETCHCONTENT_FULLY_DISCONNECTED=ON for your local build.

qualcomm-software/docs/building.md
jonathonpenix, line 53: Why remove this?

I guess it should be NFC given the current defaults, but seems unrelated to this patch.

`apply` is important here I think, so I don't think it is necessarily an issue to be explicit about that

qualcomm-software/docs/developing.md
jonathonpenix: The patching script can use `am`, but doesn't always (and not by default/in most of our current uses).

I think that part can be removed

qualcomm-software/cmake/patch_repo.py
jonathonpenix, line 28: Is there a limitation that this can't happen with `am`?

jonathonpenix, line 117: Why this change?

jonathonpenix, line 144: Similar to above, do we actually care about `capture_output`/`text`?

jonathonpenix (commented):
I agree that the current situation can be a bit painful, and I think something like this makes sense to me?

The one thing I'm unsure about is ex: if we have a patch that is essentially an "early merge" from upstream (while it is still under review, etc.) and that patch eventually comes in from upstream would this treat the patch as already applied?

I wonder if it'd be helpful to error in that case (though arguably people should keep an eye on when their patches will come in). I don't think we could actually distinguish the two cases either. I think this is only a minor concern though.

qualcomm-software/cmake/patch_repo.py
bojle: No limitation, I added support for just git apply as I am unsure why we have two variants to do the same thing.

qualcomm-software/docs/building.md
bojle: I removed it as patch_repo.py uses 'apply' as the default method, and providing `--method apply` is just an affirmation. Thought it might make the command line cleaner (fewer flags to remember). That said, I don't have a problem with explicitly providing `--method`.

qualcomm-software/cmake/patch_repo.py
bojle: Both can be removed, they are remnants from testing. Although, capture_output can be useful as without it, git will emit its error messages to the stdout when forward and reverse check are run (either one is bound to fail atleast once) - capture_output keeps it clean.

qualcomm-software/docs/building.md
jonathonpenix: I get what you mean about making it cleaner.

Just to expand on above a bit, I guess my hesitation about just leaving it to the script default is that I think `apply` is _the_ correct option for this use case given ex: how we generate VERSION.txt, see [generate_version_txt.cmake](https://github.com/qualcomm/cpullvm-toolchain/blob/qualcomm-software/qualcomm-software/cmake/generate_version_txt.cmake)).

Hopefully it can _maybe_ save someone from some confusion if they read "no explicit `--method`" as "either `apply` or `am` works here" and get confusing SHAs in their VERSION.txt.

qualcomm-software/cmake/patch_repo.py
jonathonpenix: This is code we inherited from Arm--I didn't check their original commits or anything, but as I see it:
- `apply` is helpful for our "normal" operation--we can patch the repos and check the git revision after the fact to generate VERSION.txt, etc. See [generate_version_txt.cmake](https://github.com/qualcomm/cpullvm-toolchain/blob/qualcomm-software/qualcomm-software/cmake/generate_version_txt.cmake)
- `am` is helpful when you need to work on that repo--so ex: put the patches into the repo and create or cherry-pick more commits on top that you want to create patchfiles of, rebase the patch files, etc.

jonathonpenix: In the "normal" path where it applies cleanly, there shouldn't be any error messages right (in that the reverse check wouldn't run and subsequently fail)? Seems like there wouldn't be any noise there.

Then in the "worst" case where the patch doesn't apply (so both checks would fail?) it seems like you'd want at least _some_ of git's error messages.

jonathonpenix, line 135: As-is, would this actually work if multiple patches are already applied that "interact" with each other at a text level?

Ex: if patch 1, 2, and 3 were applied in that order and they touch the same code (patch 3 tweaks code added in patch 2, which tweaks code added in patch 1, etc.):
- They should apply cleanly the first time around
- When we reapply, the (re-) application for patch 1 would fail as it doesn't apply
- Then we try to reverse patch 1 but I think that should fail too since patch 2/3 modified the code

Seems like we'd have to reverse from the top of the patch "stack" unless I'm misunderstanding something?

jonathonpenix (commented):
Sorry, revisiting this after a while.

I still agree that the current situation can be annoying to deal with, but I'm still not sure what the right path forward is.

Ultimately, I think changes like this are purely for developers (since like Navaneeth mentioned the buildbot builds should always be fresh) but getting this to work robustly in those cases seems tricky without knowing what people are doing?

Ex: I don't think it is unreasonable to imagine someone checking out their own repos, trying to patch them manually, and _not_ wanting the reverse checks -> patch application to be skipped.

Maybe just making it easier to reset the repos before applying patches again would be sufficient? It might be simpler/more "safe" in that it'd be something the developer still has to do themselves, so there isn't any guesswork on our part about intent.

I'd be curious to hear your thoughts on this stuff/if you still are interested in a change like this, etc.

PR #352, opened 2026-05-11
[cpullvm] Add armv7m -mfpu=fpv5-d16 hard float variant

jonathonpenix:
The intention here is to offer a variant for Cortex M7 users with fpv5-d16 enabled.

There's a few misc. things to note here:
- The normalized triple clang uses when `-mcpu=cortex-m7` is used is actually along the lines of `thumbv7em*` rather than the `thumb7m` we use here. Our `multilib.yaml.in` has mappings for these which we inherited from ATfE. So, align with what their configs seem to do and stick with thumbv7m.
- Picolibc is the only libc we want to support for this variant--a musl-embedded variant is intentionally omitted as we're moving away from musl-embedded.
- This variant is built without PIC support. The relevant user isn't using PIC and we've had issues with PIC + eld + Zephyr in the past for 32bit Arm targets (see https://github.com/qualcomm/cpullvm-toolchain/pull/309) so let's default towards nopic.
- We build our other armv7m configurations with `-mno-unaligned-access` (and we don't have both aligned and unaligned variants like ATfE). But, the relevant user build here is resolving to `-munaligned-access` so I'm not going to force `-mno-unaligned-access` in the library build unless we have a particular need.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -58,7 +58,7 @@
         {
             "variant": "armv7m_hard_fpv5_d16_nopic",
             "json": "armv7m_hard_fpv5_d16_nopic.json",
-            "flags": "--target=thumbv7m-unknown-none-eabihf -mfpu=fpv5-d16 -fno-pic",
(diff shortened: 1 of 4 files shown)


jonathonpenix:
Version of https://github.com/qualcomm/cpullvm-toolchain/pull/351 in the 22.x branch

apazos (approved):

PR #355, opened 2026-05-13
[cpullvm] Add missing basic multilib tests, update armv8 multilib triple

jonathonpenix:
A few variants were missing tests checking that their multilib was being picked up appropriately:
- armv8_soft_neon
- armv7m_soft_nofp
- armv7m_soft_nofp_nopic
- riscv32gc_ilp32d

Add a few very basic tests to give at least *some* coverage that we're selecting the right variant.

As part of this, update the triple we match against for our armv8_soft_neon variant. Apparently clang will normalize these flags to `thumbv8a` (rather than `thumbv8`) so we couldn't really produce a match for this multilib variant.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -38,7 +38,7 @@
         {
             "variant": "armv8_soft_neon",
             "json": "armv8_soft_neon.json",
-            "flags": "--target=thumbv8-unknown-none-eabi -mfpu=neon-fp-armv8"
+            "flags": "--target=thumbv8a-unknown-none-eabi -mfpu=neon-fp-armv8"
         },
         {
             "variant": "armv7a_soft_neon",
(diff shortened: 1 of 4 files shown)


jonathonpenix:
Note this should conflict with https://github.com/qualcomm/cpullvm-toolchain/pull/351 which also adds armv7m.test, but shouldn't matter which one goes in first.

apazos (approved):

pzhengqc (approved):

zhaoshiz (approved):
Thanks for adding these missing tests!

jonathonpenix:
Sorry, needed to fix some silly mistakes on the tests on my end (I should have actually run these I guess 😅 )

One interesting change here is apparently clang uses a normalized triple of `thumbv8a-unknown-none-eabi` for `clang --target=arm-none-eabi -march=armv8a -mfloat-abi=softfp -mthumb -mfpu=neon-fp-armv8 ...` _not_ `thumbv8-unknown-none-eabi`. Not sure why this is (for ex: armv7a + thumb it goes to `thumbv7-*`) but fixing this up here too.

apazos:
Thanks for fixing this @jonathonpenix 
Can you check if we have multilib variants/tests for other Arm v8 profiles other than A,  like M and R, that would require similar changes?

jonathonpenix:
After these changes, we should have (at least) one test per variant.

So I think we're covered/there shouldn't be more similar changes to the `thumbv8a` thing. (Unless our tests are broken somehow, but I think we're ok.)

jonathonpenix:
Discussed with Ana, merging.

PR #357, opened 2026-05-14
[cpullvm] Force a consistent TLS model for all library builds using picolibc

jonathonpenix:
Previously, a TLS model was only ever explicitly set for picolibc--other projects (libc++, etc.) would use whatever the compiler selected. This shouldn't necessarily be a problem, but in practice can cause issues when global-dynamic is selected and is implemented via calls to `__tls_get_addr` (ex: in our 32bit Arm and 32/64bit RISC-V PIC configs). Given we don't provide an implementation and we don't expect people to use a dynamic linker that would implement this, people will generally get undefined reference errors. See the discussion in https://github.com/qualcomm/cpullvm-toolchain/pull/341.

Note that this seems to primarily be an issue with picolibc as it uses "normal" TLS mechanisms [1]. musl-embedded doesn't seem to (see ex: `__errno_location`) and consequently I don't see any obvious TLS usage in any of our libraries built with musl-embedded as the libc.

To try and address this, consistently specify a TLS model across all of the libraries we are building when using picolibc. 

For musl-embedded, given that a) it is generally not expected to be thread safe b) it doesn't seem to use the "usual" TLS mechanisms from the toolchain as mentioned above and c) historically we haven't had any special handling downstream for this, musl-embedded configs are intentionally excluded here.

Since we're special casing the flag on the libc type now, at the same time add basic error checking so we aren't silently dropping this if it is specified via JSON.

[1] https://github.com/picolibc/picolibc/blob/main/doc/tls.md

### DIFF
--- a/qualcomm-software/embedded-runtimes/CMakeLists.txt
+++ b/qualcomm-software/embedded-runtimes/CMakeLists.txt
@@ -88,11 +88,13 @@ set(DISABLE_THREADS ${DISABLE_THREADS_def} CACHE BOOL "Disable threads/atomic/TL
 set(SUPPORTED_TLS_MODELS local-exec initial-exec local-dynamic global-dynamic)
 if(NOT DEFINED TLS_MODEL_def)
     set(TLS_MODEL_def local-exec)
+elseif(C_LIBRARY STREQUAL musl-embedded)
+    message(FATAL_ERROR "TLS_MODEL is currently unsupported when using musl-embedded.")
 elseif(NOT TLS_MODEL_def IN_LIST SUPPORTED_TLS_MODELS)
     string(REPLACE ";" ", " SUPPORTED_TLS_MODELS_STR "${SUPPORTED_TLS_MODELS}")
     message(FATAL_ERROR "Unsupported TLS_MODEL value. Supported values are ${SUPPORTED_TLS_MODELS_STR}")
 endif()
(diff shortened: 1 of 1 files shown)


jonathonpenix:
CC'ing @bojle since I can't add you as a reviewer directly (I think you need to join the Qualcomm Github org before we can?)

apazos:
Our musl-embedded is not thread safe. We can ignore it.

qualcomm-software/embedded-multilib/json/variants/aarch64a_tlsie.json
```diff
@@ -19,8 +20,7 @@
             "ENABLE_CXX_LIBS": "ON",
             "ENABLE_LIBC_TESTS": "ON",
             "ENABLE_COMPILER_RT_TESTS": "OFF",
-            "ENABLE_LIBCXX_TESTS": "OFF",
-            "TLS_MODEL": "initial-exec"
+            "ENABLE_LIBCXX_TESTS": "OFF"
         },
         "musl-embedded": {
```
apazos, line 25: let's move musl-embedded to a separate config where no TLS model is defined.

bojle:
Looks good to me, I'll try rebasing this patch and testing the libcxxabi patch with it

apazos:
Thanks @jonathonpenix , we are fine with excluding the deprecated musl-embedded.

apazos (approved):

PR #360, opened 2026-05-15
[cpullvm] Don't stop on first test failure in Linux builds, write test results to xml logs

jonathonpenix:
With automerge now enabled, we have had more consistent failures, even in Linux x86 builds. And, we've now seen failures that have been masked by other failures since we stop after the first one.

So let's try what seems to be done in ATfE and use lit's `--ignore-fail` and dump all results into xml.

A few things to note:
1. Windows builds are not impacted by this--they will still fail at the first error. We don't have QEMU testing enabled there, so we're really only running the lit tests anyway, so I don't think this is as large of an issue. We can re-evaluate this as needed.
2. We might need to think about how this works for ex: PRs. As-is, if there are failures I *think* the precheckin build will still succeed. So either we need an extra step, or people need to be diligent about checking the test reports.
3. I'm also not sure how this will work with eld? Given that its lit tests are split with different defaults. There might be some work needed to make this work well.

Note that the code added here is from or is derived from ATfE. All license/copyright info we already have in the modified files should already be correct/appropriate.

### DIFF
--- a/.github/workflows/native-runtime-nightly.yml
+++ b/.github/workflows/native-runtime-nightly.yml
@@ -104,3 +104,10 @@ jobs:
       - name: Test
         if: matrix.test_script != ''
         run: ./qualcomm-software/scripts/${{ matrix.test_script }}
+
+      - name: Upload test results
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+        if: always()
+        with:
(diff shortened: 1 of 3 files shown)


jonathonpenix:
This might need more work as mentioned in the description, but just creating this for discussion about what Arm does, to see what the results would look like, etc.

Edit: Also confirming that we're going to need to do something for eld to handle the whole "invoke lit multiple times" thing

PR #362, opened 2026-05-18
[cpullvm] use apt-get -y and add libc++ dependencies

navaneethshan:
These packages are required for ELD unit tests

### DIFF
--- a/qualcomm-software/scripts/install_dependencies.sh
+++ b/qualcomm-software/scripts/install_dependencies.sh
@@ -5,8 +5,8 @@ sudo apt-get update
 # Used by lldb
 sudo apt-get install -y swig libedit-dev
 
-# Install clang, libc++ and libc++abi
-sudo apt-get install -y clang libc++-dev libc++abi-dev libclang-rt-20-dev
+# Install libc++ and libc++abi required for eld tests
+sudo apt-get install -y libc++-dev libc++abi-dev
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
 pip install meson==1.10.0
--- a/qualcomm-software/scripts/install_dependencies.sh
+++ b/qualcomm-software/scripts/install_dependencies.sh
@@ -4,7 +4,11 @@
 sudo apt-get update
 # Install swig and libedit-dev required for lldb and
 # libc++-dev required for eld tests
-sudo apt-get install -y swig libedit-dev libc++-dev
+sudo apt-get install -y swig libedit-dev clang-19 libc++-19-dev
+
+# Ensure system uses clang-19
+sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-19 100
+sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-19 100
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
 pip install meson==1.10.0
--- a/qualcomm-software/scripts/install_dependencies.sh
+++ b/qualcomm-software/scripts/install_dependencies.sh
@@ -3,7 +3,10 @@
 
 sudo apt-get update
 # Used by lldb
-sudo apt-get install swig libedit-dev
+sudo apt-get install -y swig libedit-dev
+
+# Install clang, libc++ and libc++abi
+sudo apt-get install -y clang libc++-dev libc++abi-dev libclang-rt-20-dev
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
 pip install meson==1.10.0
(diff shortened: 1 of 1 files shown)


navaneethshan:
Verifying "Linux-AArch64 with the Nightly build" on this PR/branch: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/26049751406

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -3,7 +3,10 @@
 
 sudo apt-get update
 # Used by lldb
-sudo apt-get install swig libedit-dev
+sudo apt-get install -y swig libedit-dev
+
+# Install clang, libc++ and libc++abi
+sudo apt-get install -y clang libc++-dev libc++abi-dev libclang-rt-20-dev
```
jonathonpenix: Re: `clang`: clang is already installed on the Github machines we use. When we do this, are we actually going to get whatever clang we install here? Is this going to implicitly change the version of clang we are building with?

Re: `libclang-rt-20-dev`: is this actually required? libc++ is required since they apparently do an explicit -stdlib=libc++ (which, I'm not convinced is a good idea, but w/e), but I really hope they don't also do with compiler-rt.

Also, the version of compiler-rt here must match whatever version of clang we use. If we have a hardcoded compiler-rt here, I think we also have to be very careful about making sure we aren't picking up something other than clang-20.

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -3,7 +3,10 @@
 
 sudo apt-get update
 # Used by lldb
-sudo apt-get install swig libedit-dev
+sudo apt-get install -y swig libedit-dev
+
+# Install clang, libc++ and libc++abi
```
jonathonpenix: Can you mention these are required for eld tests currently?

navaneethshan: Done

qualcomm-software/scripts/install_dependencies.sh
navaneethshan: Ah, yes, eld did some additional changes to install clang version 20, which we don't have to do now.
Added only libc++-dev and libc++abi-dev, verifying it.

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -3,7 +3,10 @@
 
 sudo apt-get update
 # Used by lldb
-sudo apt-get install swig libedit-dev
+sudo apt-get install -y swig libedit-dev
+
+# Install libc++ and libc++abi required for eld tests
+sudo apt-get install -y libc++-dev libc++abi-dev
```
apazos: Let's keep the dependencies to the minimal required to run ELD tests which is only libc++-dev

navaneethshan: OK.

Testing it with https://github.com/qualcomm/cpullvm-toolchain/actions/runs/26102294978

navaneethshan:
Default clang version on Ubuntu24.04 on GH hosted machines is clang-18: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

AArch64 test verified with this PR, its passing: https://github.com/qualcomm/cpullvm-toolchain/pull/362/changes

jonathonpenix:
Thanks for looking into this!

This seems ok to me provided we're ok with bumping the version of clang we're using (presumably this means we're also using 19 for the main toolchain build vs 18). IIRC we're using 19 on x86 Linux so matching that version seems fine to me. 

Thoughts/preferences on this @apazos?

If clang is looking for a libc++ of a particular version (I'm assuming that'd come down to where it gets installed?) presumably we could also use libc++ 18 if we didn't want to switch clang versions?

If we update the clang version, please do mention that in the final commit message when merging.

apazos:
Thanks @navaneethshan and @jonathonpenix, I think it is fine to move to clang 19, as part of host toolchain unification we want to update the host clang for all 4 builders, at least with this change 2 builders are aligned.

apazos (approved):

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -2,8 +2,13 @@
...
 sudo apt-get update
-# Used by lldb
-sudo apt-get install swig libedit-dev
+# Install swig and libedit-dev used by lldb and
+# libc++-dev required for eld tests
+sudo apt-get install -y swig libedit-dev clang-19 libc++-19-dev
+
+# Set default clang version to clang-19
+sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-19 100
+sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-19 100
```
apazos, line 11: Mention in commit message you upgraded to clang 19 for Linux on Arm64 builder.

apazos (approved):

PR #367, opened 2026-05-19
[cpullvm] Add missing basic multilib tests, update armv8 multilib triple

jonathonpenix:
A few variants were missing tests checking that their multilib was being picked up appropriately:
- armv8_soft_neon
- armv7m_soft_nofp

Add a few very basic tests to give at least *some* coverage that we're selecting the right variant.

As part of this, update the triple we match against for our armv8_soft_neon variant. Apparently clang will normalize these flags to thumbv8a (rather than thumbv8) so we couldn't really produce a match for this multilib variant.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -38,7 +38,7 @@
         {
             "variant": "armv8_soft_neon",
             "json": "armv8_soft_neon.json",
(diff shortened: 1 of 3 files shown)


jonathonpenix:
This is essentially a cherry-pick of https://github.com/qualcomm/cpullvm-toolchain/pull/355 minus the configs that aren't present in the branch.

apazos (approved):

apazos (approved):

PR #372, opened 2026-05-20
[cpullvm] Create separate non-'tlsie' musl-embedded configs

jonathonpenix:
The naming here is confusing as we don't do anything related to TLS models for the musl-embedded configs. So, split these apart.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -1,9 +1,16 @@
 {
     "libs": [
         {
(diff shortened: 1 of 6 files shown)


jonathonpenix:
FWIW, one unfortunate thing here is that, as-is, we don't really have a setup for testing non-picolibc multilib matching. In theory everything should be identical modulo the variant name, but 🤷

apazos:
It is ok if we do not test musl-embedded since we are deprecating it.

apazos (approved):

PR #373, opened 2026-05-20
[cpullvm] Added build-essential to restore missing C++ headers on ubu…

pranav4330:
Ubuntu 24.04 ARM runner no longer includes default C++ headers (libstdc++-dev).
This causes missing header errors like <cstdint> during builds.
Adding build-essential restores the standard C++ toolchain and fixes the issue.

### DIFF
--- a/qualcomm-software/scripts/install_dependencies.sh
+++ b/qualcomm-software/scripts/install_dependencies.sh
@@ -3,7 +3,7 @@
 
 sudo apt-get update
 # Used by lldb
-sudo apt-get install swig libedit-dev
+sudo apt-get install swig libedit-dev build-essential
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
 pip install meson==1.10.0


qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -3,7 +3,7 @@
 
 sudo apt-get update
 # Used by lldb
-sudo apt-get install swig libedit-dev
+sudo apt-get install swig libedit-dev build-essential
```
apazos, line 7: build-essential brings GCC and GCC libraries, no LLVM libraries which is the cause of builder failure.

apazos: @pranav4330, 

Go back to the code that installs:
sudo apt install -y libc++-dev libc++abi-dev 

And then try to run these commands after the installation to understand better what is failing:
echo '#include <iostream>' > test.cpp
          echo 'int main() { std::cout << "Hello, Clang on Ubuntu 24.04!" << std::endl; return 0; }' >> test.cpp
          clang++ -std=c++17 -stdlib=libc++ test.cpp -o test_clang
          ./test_clang

If this test passes, then you successfully resolved the problem. Otherwise, it might give us more hints on what needs to be fixed.

apazos: You can also add -v to clang command line to see where clang is searching for C++ lib, and see if it matches what you installed.

pranav4330: Okay Thank you for the input, i will check on this once and update.

jonathonpenix:
I think this has been superseded by https://github.com/qualcomm/cpullvm-toolchain/pull/362 right?

Might want to close this one/mark it as a draft just to avoid confusion

jonathonpenix (changes requested):
Setting "Request changes" just to mark that #362 is the active review to fix this

pranav4330:
Alternative PR added by Navneeth and Merged.

PR #374, opened 2026-05-21
[cpullvm] Upgrade ram sizes for large test cases (for ARM/AArch64)

bojle:
Many tests such as dynamic_cast14.pass.cpp in libcxxabi fail as their compiled binary does not fit in the RAM size set for qemu. This patch upgrades the RAM size of qemu configurations for each variant impacted (in this case, its all the variants). This PR deals with just ARM/AArch64 variants. Riscv variants are handled in https://github.com/qualcomm/cpullvm-toolchain/pull/364

RAM sizes directly from https://github.com/arm/arm-toolchain/tree/arm-software/arm-software/embedded/arm-multilib/json/variants

PR #375, opened 2026-05-21
[cpullvm] upgrade ram size for riscv variants

bojle:
Tests such as dynamic_cast14.pass.cpp in libcxxabi fail due to
infsufficient RAM size. This patch fixes it by increasing the RAM size
allocated to qemu.

Tests such as dynamic_cast14.pass.cpp in libcxxabi fail due to
infsufficient RAM size. This patch fixes it by increasing the RAM size
allocated to qemu. The sizes are derived from the reports of failed test
cases empirically. dynamic_cast14 overflows by 7.7MB, its executable
size is 2MB, which implies a requirement of ~9.5MB. Therefore, a size
of 10MB is chosen for all riscv variants. Other tests, as they're
uncovered may demand a higher size, in which case this can be adjusted.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/variants/riscv32gc_ilp32d.json
+++ b/qualcomm-software/embedded-multilib/json/variants/riscv32gc_ilp32d.json
@@ -13,7 +13,7 @@
             "FLASH_ADDRESS": "0x80000000",
             "FLASH_SIZE": "0x00400000",
             "RAM_ADDRESS": "0x80400000",
-            "RAM_SIZE": "0x00200000",
+            "RAM_SIZE": "0x00a00000",
             "LIBRARY_BUILD_TYPE": "minsizerelease"
         },
         "picolibc": {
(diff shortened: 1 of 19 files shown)


bojle:
related https://github.com/qualcomm/cpullvm-toolchain/pull/374

jonathonpenix:
Can you include a mention in the commit message/PR description about where the new size is coming from (is it just a value that was large enough experimentally for the tests in question?) would be helpful.

bojle:
@jonathonpenix I've updated the description with a reference to where the sizes come from.

apazos:
@bojle, can you add the information about how you adjusted the sizes to the commit message? It is helpful for future maintainers. @quic-garvgupt will do a combo run with his additional patches to confirm the issues are resolved.

bojle:
Added the method of deciding on the ram size to the commit message and the PR description. @apazos please check it out.

apazos (approved):

jonathonpenix (approved):

apazos (approved):

PR #389, opened 2026-05-26
[cpullvm] Specify and require `-munaligned-access` for Arm/AArch64 variants that currently build with unaligned access

### LINKED ISSUE #388: Decide how to handle alignment requirements in Arm/AArch64 configs
Currently, our Arm/AArch64 variants have a mix in terms of how they deal with alignment requirements that I think we might want to iron out/have room to improve upon.

A quick attempt at summarizing the situation is below:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/jpenix/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/jpenix/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">

Variant | Alignment flag picolibc | Flag impacts other libs? | Alignment flag musl-embedded | Flag impacts other libs? | Alignment enforced by multilib? | clang default alignment flag for   multilib using compile flags
-- | -- | -- | -- | -- | -- | --
aarch64a_tlsie | No | N/A | -mstrict-align | No | No | munaligned-access
aarch64a_pacret | No | N/A | -mstrict-align | No | No | munaligned-access
aarch64a_pacret_bti | No | N/A | -mstrict-align | No | No | munaligned-access
aarch64a_pacret_bkey_bti_tlsie | No | N/A | -mstrict-align | No | No | munaligned-access
aarch64a_soft_nofp_tlsie | No | N/A | -mstrict-align | No | No | munaligned-access
aarch64a_soft_nofp_pacret_bti | No | N/A | -mstrict-align | No | No | munaligned-access
aarch64a_soft_nofp_pacret_bkey_bti | No | N/A | -mstrict-align | No | No | munaligned-access
armv8_soft_neon | No | N/A | -mno-unaligned-access | No | No | munaligned-access
armv7a_soft_neon | No | N/A | -mno-unaligned-access | No | No | munaligned-access
armv7a_soft_nofp | No | N/A | -mno-unaligned-access | No | No | munaligned-access
armv7m_soft_nofp | -mno-unaligned-access | Yes | -mno-unaligned-access | Yes | No | munaligned-access (-mno- given the compile flag)
armv7m_soft_nofp_nopic | -mno-unaligned-access | Yes | N/A | N/A | No | munaligned-access (-mno- given the compile flag)
armv7m_hard_fpv5_d16_nopic* | No | N/A | N/A | N/A | Yes | munaligned-access

</body>

</html>

*Not yet merged, but as proposed in #351

Takeaways for picolibc-based variants:

- All variants are compiled with unaligned accesses enabled and no multilib "guards" except armv7m_soft_nofp, armv7m_soft_nofp_nopic, and armv7m_hard_fpv5_d16_nopic
- armv7m_soft_nofp and armv7m_soft_nofp_nopic are compiled with `-mno-unaligned-access` (all libraries) and have no special multilib matching.
    - IIRC, this is just matching what we did downstream
- armv7m_hard_fpv5_d16_nopic is compiled with unaligned access but does require `-munaligned-access` for multilib matching.
    - See the discussion in #351 for why this happened.

I think we can have issues here in that, for all variants in the first category, if someone specifies -mno-unaligned-access (or similar), they'll silently get an unaligned library build. If their hardware doesn't support unaligned accesses (I'm assuming some implementations could have this turned off or unimplemented?), that seems bad. Even if it is supported, it might not be ideal I guess.

From #351, it seems like there's three safe(-ish?) ways we can handle this:
1. Start matching -munaligned-access (or equivalent) in our multilibs and leave our build flags as-is. This way we at least have some compile-time safeguards in place. Seems like this might cause problems if aligned accesses are more of a preference rather than a requirement? Not sure how often that case arises/what the impacts are here.
2. Leave our multilib matching as-is, but more consistently force alignment (set -mno-unaligned-access or equivalent) in the build flags. I think this should be safer that what we have today? Not sure if there are other impacts (code size, performance, etc.).
3. Provide both aligned and unaligned configs similar to Arm.

jonathonpenix:
Requiring `-munaligned-access` for multilib should provide better build-time
safeguards if people try using `-mno-unaligned-access` when they truly need it
disabled.

If people require an aligned variant, we'll consider adding one in the future.
And, if people want to mix and match (build with `-mno-unaligned-access` but
match against a library built with `-munaligned-access`) they can still
override multilib and point at the library variant manually.

At the same time, expicitly specify `-munaligned-access` for these configs in
the compile flags. This should be NFC, but will hopefully make it more clear
what our intent for these variants is and safeguard us from any unforseen
changes in defaults, etc.

Note that this also shouldn't impact our musl configs where we specify
`-mstrict-align`/`-mno-unaligned-access` via
`EXTRA_MUSL-EMBEDDED_CFLAGS`--those are used when building musl *only* and
those flags are placed after the generic `COMPILE_FLAGS` so will override them.

As mentioned above, this should resolve https://github.com/qualcomm/cpullvm-toolchain/issues/388
for now until we want to support aligned variants.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -3,73 +3,73 @@
         {
             "variant": "aarch64a",
             "json": "aarch64a.json",
-            "flags": "--target=aarch64-unknown-none-elf",
+            "flags": "--target=aarch64-unknown-none-elf -munaligned-access",
             "libraries_supported": "musl-embedded"
         },
         {
             "variant": "aarch64a_tlsie",
             "json": "aarch64a_tlsie.json",
-            "flags": "--target=aarch64-unknown-none-elf",
+            "flags": "--target=aarch64-unknown-none-elf -munaligned-access",
             "libraries_supported": "picolibc"
         },
         {
             "variant": "aarch64a_pacret",
             "json": "aarch64a_pacret.json",
-            "flags": "--target=aarch64-unknown-none-elf -march=armv8.3-a -mbranch-protection=pac-ret+leaf"
+            "flags": "--target=aarch64-unknown-none-elf -march=armv8.3-a -mbranch-protection=pac-ret+leaf -munaligned-access"
         },
         {
             "variant": "aarch64a_pacret_bti",
             "json": "aarch64a_pacret_bti.json",
-            "flags": "--target=aarch64-unknown-none-elf -march=armv8.5-a -mbranch-protection=pac-ret+leaf+bti"
+            "flags": "--target=aarch64-unknown-none-elf -march=armv8.5-a -mbranch-protection=pac-ret+leaf+bti -munaligned-access"
         },
         {
             "variant": "aarch64a_pacret_bkey_bti",
             "json": "aarch64a_pacret_bkey_bti.json",
-            "flags": "--target=aarch64-unknown-none-elf -march=armv8.5-a -mbranch-protection=pac-ret+leaf+b-key+bti",
+            "flags": "--target=aarch64-unknown-none-elf -march=armv8.5-a -mbranch-protection=pac-ret+leaf+b-key+bti -munaligned-access",
             "libraries_supported": "musl-embedded"
         },
         {
             "variant": "aarch64a_pacret_bkey_bti_tlsie",
             "json": "aarch64a_pacret_bkey_bti_tlsie.json",
-            "flags": "--target=aarch64-unknown-none-elf -march=armv8.5-a -mbranch-protection=pac-ret+leaf+b-key+bti",
+            "flags": "--target=aarch64-unknown-none-elf -march=armv8.5-a -mbranch-protection=pac-ret+leaf+b-key+bti -munaligned-access",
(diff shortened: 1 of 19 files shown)


apazos (commented):
- Let's add a test case where we enable -munaligned-access explicitly
- It is good that we explicitly build with -munaligned-access instead of relying on clang default behavior for arm target

jonathonpenix:
Rebased to fix the Linux failures.

> Let's add a test case where we enable -munaligned-access explicitly

Sprinkled in a few (non-exhaustive) tests to hopefully give some coverage of this.

apazos (approved):

zhaoshiz (approved):

PR #392, opened 2026-05-28
Workflow to build 22.x on ubuntu24 OS

sriyalamar:
Change-Id: Id44c23c392208bca94930e29346e47f635b2a9bf

### DIFF
--- a/.github/workflows/nightly-ubuntu24.yml
+++ b/.github/workflows/nightly-ubuntu24.yml
@@ -1,0 +1,139 @@
+# Copyright (c) 2025, Arm Limited and affiliates.
+# Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# ​​​​​Changes from Qualcomm Technologies, Inc. are provided under the following license:
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This workflow runs our common build scripts in parallel for multiple operating
+# systems and architectures. All runtimes are built on Linux x86_64 and copied to
+# other build hosts.
+#
+# It is intended to be triggered as part of a **nightly build** to
+# validate our typical build configurations and test them automatically.
+#
+# Other build configurations such as building the runtimes on non-Linux
+# x86_64 platforms are built and tested separately.
+
+name: Nightly Build in Ubuntu24 and Test
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  build-and-test-Linux-x86:
+    # Don't run on forks
+    if: github.repository == 'qualcomm/cpullvm-toolchain'
+    name: build.sh on Linux-x86_64
+    runs-on: cpullvm-ubuntu24-x86_64
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          # We build in the repository and rely on clean to cleanup possible
+          # old build products if the runner doesn't do it itself.
+          clean: true
+
+      - name: Apply llvm-project patches
+        run: python3 qualcomm-software/cmake/patch_repo.py --method apply qualcomm-software/patches/llvm-project
+
+      - name: Build
+        run: ./qualcomm-software/scripts/build.sh
+
+      - name: Upload built packages
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+        if: success()
+        with:
+          name: cpullvm-packages-build.sh-Linux-x86
+          path: build*/cpullvm-*.tar.xz
+          if-no-files-found: error
+          retention-days: 7
+
+      - name: Test
+        run: ./qualcomm-software/scripts/test.sh
+
+  build-and-test:
+    if: github.repository == 'qualcomm/cpullvm-toolchain'
+    name: ${{ matrix.build_script }} on ${{ matrix.target_os }}
+    needs: build-and-test-Linux-x86
+    runs-on: ${{ matrix.runner }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - build_script: build_copy_runtimes.sh
+            test_script: test.sh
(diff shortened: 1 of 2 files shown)


qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -2,8 +2,13 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
jonathonpenix, line 2: We probably shouldn't make changes here in the branch--we presumably want our Ubuntu 22 builds to be untouched, and this is shared with those.

A separate script for Ubuntu 24 seems like it would work.

jonathonpenix (changes requested):
How do we want Ubuntu 24 builds to work?

Given what we released for 22.1.1, it seems like we should just add new Ubuntu 24 entries in the matrix in the normal nightly.yml (and maybe the native runtimes one too).

Or do we actually need/want a full set of new Windows builds, etc. with different runtimes?

Edit: maybe ignore this, see comment below.

jonathonpenix:
> Given what we released for 22.1.1...

Assuming we used the earlier revision of this to make the builds, I guess the Ubuntu 24 builds are also using a different set of runtimes than the others so adding to the matrix doesn't really replicate what happened in the release.

So I guess my question is really more whether we want the Windows builds or not in this workflow? AFAICT we didn't release something like that earlier, do we have plans to?

jonathonpenix:
Sorry, two more asks/questions:
1. Doesn't need to be right now, but once the code starts to settle, please do a run with these changes to show the nightly.yml + Ubuntu 24 nightly working as expected after the updates and post the links.
2. Should we be adding a Ubuntu 24 build into the precheckin on the branch? Given that we're going to release Ubuntu 24 builds going forward (AFAICT) and the builds should be fairly quick, it might be worth trying to catch issues earlier rather than just in the nightly.

apazos:
Confirming that we do not do replicate window builds.

jonathonpenix:
Discussed with @apazos and @Yethendra, this is my understanding of where we landed:
1. The separate workflows like what you have is ok, let's just remove musl-embedded and the Windows builds from the Ubuntu 24 workflows. Longer term we can see if there's another way to support this.</br></br>I think the immediate issue with the matrix stuff is we'd either need to a) have the Ubuntu 24 builds use the Ubuntu 22 runtimes (which, might be ok but isn't what we initially released) b) build the runtimes on Ubuntu 24 AArch64 natively (which again might be ok long term but isn't what we've done so far) or c) rework the nightly itself to support sharing runtimes between "similar" Ubuntu versions, which seems like it might be more complicated than it is worth.
2. I think we also should start automatically including the Ubuntu versions in the file names (for Ubuntu 24 in the branch, and all Ubuntu versions in mainline). I left comments above about doing this.

.github/workflows/nightly-ubuntu24.yml
```diff
@@ -0,0 +1,139 @@
...
+    runs-on: ${{ matrix.runner }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - build_script: build_copy_runtimes.sh
+            test_script: test.sh
+            install_script: install_dependencies.sh
+            target_os: Linux-AArch64
```
jonathonpenix, line 73: Can you update this to include `Ubuntu24` somewhere (or whatever our naming convention is going to be)

.github/workflows/nightly-ubuntu24.yml
```diff
@@ -0,0 +1,139 @@
...
+        run: python3 qualcomm-software/cmake/patch_repo.py --method apply qualcomm-software/patches/llvm-project
+
+      - name: Build
+        run: ./qualcomm-software/scripts/build.sh
+
+      - name: Upload built packages
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+        if: success()
+        with:
+          name: cpullvm-packages-build.sh-Linux-x86
```
jonathonpenix, line 52: Include Ubuntu24 indication here

.github/workflows/nightly-ubuntu24.yml
```diff
@@ -0,0 +1,139 @@
...
+
+      - name: Upload built packages
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+        if: success()
+        with:
+          name: cpullvm-packages-${{ matrix.build_script }}-${{ matrix.target_os }}
+          path: |
+            build*/cpullvm-*.tar.xz
+            build*/cpullvm-*.zip
+            !build/cpullvm-*-Linux-x86_64.tar.xz
```
jonathonpenix, line 134: You'll need to double check, but I think this will need to be updated too so we don't reupload the toolchain from the first step again.

.github/workflows/nightly-ubuntu24.yml
```diff
@@ -0,0 +1,139 @@
...
+        with:
+          # We build in the repository and rely on clean to cleanup possible
+          # old build products if the runner doesn't do it itself.
+          clean: true
+
+      - name: Apply llvm-project patches
+        run: python3 qualcomm-software/cmake/patch_repo.py --method apply qualcomm-software/patches/llvm-project
+
+      - name: Build
+        run: ./qualcomm-software/scripts/build.sh
```
jonathonpenix, line 46: To rename the "real" package (not the outer zip file you can download from the action but the "inner" package containing the toolchain) I think we have a few options:
1. Set `-DPACKAGE_SYSTEM_NAME=<name>`, see [here](https://github.com/qualcomm/cpullvm-toolchain/blob/e8edac55bf4b2abc0724728a125e6b2cdd41a41c/qualcomm-software/CMakeLists.txt#L445).
2. Modify the CMakeLists.txt itself (again somewhere around [here](https://github.com/qualcomm/cpullvm-toolchain/blob/e8edac55bf4b2abc0724728a125e6b2cdd41a41c/qualcomm-software/CMakeLists.txt#L436-L451)) to try and check if we're on Ubuntu and, if so, append the version.

To me, I think 1 would be ok in the branch (so it is easier to omit Ubuntu 22), but 2 is better in mainline where we want everything labeled.

I'm not 100% sure how this will work out/how the best way to pass through -DPACKAGE_SYSTEM_NAME will be (either a new script or creating a passthrough in the matrix to feed into the passthrough of the build script, see [EXTRA_CMAKE_ARGS](https://github.com/qualcomm/cpullvm-toolchain/blob/e8edac55bf4b2abc0724728a125e6b2cdd41a41c/qualcomm-software/scripts/build.sh#L28)). You might have to experiment a bit with this.

PR #396, opened 2026-05-29
DO NOT MERGE

quic-garvgupt:
Squashed change of all the PRs required to enable compiler-rt tests for arm and riscv


quic-garvgupt:
@apazos @jonathonpenix Linux build is green. Pls note that this PR modified the test.sh script to only run compiler-rt tests

compiler-rt/test/builtins/Unit/lit.cfg.py
```diff
@@ -107,7 +107,9 @@ def get_libgcc_file_name():
     if config.target_os == "Haiku":
         config.substitutions.append(("%librt ", base_lib + " -lroot "))
     else:
-        config.substitutions.append(("%librt ", base_lib + " -lc -lm "))
+        config.substitutions.append(
```
apazos: @quic-garvgupt did you upstream this patch? We cannot have this change in LLVM tests in CPULLVM, must go through upstream.

PR #400, opened 2026-06-01
[cpullvm] Enable assertions for Release+Asserts builds

Yethendra:


apazos (approved):

qualcomm-software/CMakeLists.txt
```diff
@@ -263,13 +263,16 @@ set(CLANG_DEFAULT_LINKER lld CACHE STRING "")
...
-# Default to a release build
+# Default to Release, and enable assertions for Release+Asserts
 # (CMAKE_BUILD_TYPE is a special CMake variable so if you want to set
 # it then you have to FORCE it).
 if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE Release CACHE BOOL "" FORCE)
 endif()
 
+#Enable assertions for Release+Asserts builds
+set(LLVM_ENABLE_ASSERTIONS ON CACHE BOOL "" FORCE)
```
jonathonpenix:
```suggestion
set(LLVM_ENABLE_ASSERTIONS ON CACHE BOOL "")
```

I don't think we should be `FORCE`-ing this--if people want to set this on the command line (ex: `cmake ... -DLLVM_ENABLE_ASSERTIONS=OFF`) that should be allowed I think

(and this also lets us toggle it easily since we're also going to want builds with this turned off)

qualcomm-software/CMakeLists.txt
```diff
@@ -263,13 +263,16 @@ set(CLANG_DEFAULT_LINKER lld CACHE STRING "")
 set(LLDB_ENABLE_PYTHON OFF CACHE BOOL "")
 set(LLDB_INCLUDE_TESTS OFF CACHE BOOL "")
 
-# Default to a release build
+# Default to Release, and enable assertions for Release+Asserts
```
jonathonpenix: Maybe redundant with the comment below.

qualcomm-software/CMakeLists.txt
```diff
@@ -263,13 +263,16 @@ set(CLANG_DEFAULT_LINKER lld CACHE STRING "")
...
 # (CMAKE_BUILD_TYPE is a special CMake variable so if you want to set
 # it then you have to FORCE it).
 if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE Release CACHE BOOL "" FORCE)
 endif()
 
+#Enable assertions for Release+Asserts builds
```
jonathonpenix:
```suggestion
# Enable assertions for Release+Asserts builds
```
It might also be worth expanding this comment to say why we're enabling assertions by default

jonathonpenix:
I guess the other high level question here is whether we actually want this to be the default?

It is probably best for developers/development, but if ex: we tell someone how to build their own toolchain my guess would be that we want to disable this (or rather, strongly encourage/show them how to disable it).

apazos (changes requested):
Just noticed Jonathon's reasonable request.

jonathonpenix (approved):

apazos (approved):

PR #402, opened 2026-06-07
Add extension combo for Xqci with atomic and z*inx extensions

pgodeq:


pgodeq:
Not able to know what these errors are due to the compilicated build system. Abandoning this.

PR #404, opened 2026-06-08
[RISCV][Compiler-rt]Add libc before libsemihost to link line for RISC-V compiler-rt tests

quic-garvgupt:
Currently the link order is `... -lcrt0-semihost -lsemihost ... -lc`.
This causes unresovled errors with some riscv compiler-rt tests because
`-lcrt0-semihost` has refernces to `printf(...)` function which in turn
has references to `stdout\stderr` that are only defined in `-lsemihost`.
However `-lc` comes much after `-lsemihost` in the link line. Because of this,
for eld like linkers, at the end of link command there are unresolved references
to `stdout/stderr` symbols as eld only knows about these symbols when
it scans `-lc`.

This patch fixes this issue by adding `-lc` before `-lsemihost` only for
riscv target when QEMU testing is enabled.

This issue was not observed with arm/aarch64 because `-lcrt0-semihost` for
arm/aarch64 directly use `fputs` instead of `printf`which directly uses
`stdout/stderr` therefore there is no dependency on libc.

This issue cannot be fixed in lit.cfg file in upstream because clang
currently does not have support for adding `-lsemihost` to linker job
for baremetal targets when qemu testing is enabled and this will make
it out of sync with what is passed through config file.

Additionally, this issue is specific to picolibc and this change might not
be needed in with other libc depending on the contents of
lcrt0-semihost. Therefore fixing this in CPULLVM specific CMakeFile.

### DIFF
--- a/qualcomm-software/embedded-runtimes/CMakeLists.txt
+++ b/qualcomm-software/embedded-runtimes/CMakeLists.txt
@@ -172,7 +172,9 @@ if(ENABLE_LIBC_TESTS OR ENABLE_COMPILER_RT_TESTS OR ENABLE_LIBCXX_TESTS)
         if(TEST_EXECUTOR STREQUAL qemu)
             set(picocrt "crt0-semihost")
         endif()
-        set(test_link_flags "-fuse-ld=eld -nostartfiles -l${picocrt} -lsemihost -T picolibcpp.ld")
+        # To run compiler-rt tests on qemu, libc must be linked before
+        #  libsemihost to resolve semihost's references into libc.
+        set(test_link_flags "-fuse-ld=eld -nostartfiles -l${picocrt} -lc -lsemihost -T picolibcpp.ld")
     else()
         message(FATAL_ERROR "Tests can only be enabled using picolibc.")
     endif()


quic-garvgupt:
The build failure is unrelated and is due to some other PR for which QA has already created the superops. 

This PR is needed to enable compiler-rt tests for riscv and was demonstrated in PR #396

quic-garvgupt:
Ping!

apazos:
@quic-garvgupt the issue is only with RISC-V but not Arm?

quic-garvgupt:
> @quic-garvgupt the issue is only with RISC-V but not Arm?

yes @apazos. The same is also already mentioned in the commit msg with reasoning

quic-garvgupt:
The same issue is not observed during picolibc tests because meson wraps libsemihost under `--start/--end-group` together with libc.a and builtins. 

For reference from `picolibc/riscv32imac_ilp32/build/build.ninja`

`LINK_ARGS = -Wl,--as-needed -Wl,--no-undefined -fuse-ld=eld -Wl,--start-group newlib/libc.a semihost/libsemihost.a '-D_PICOLIBC_PRINTF='"'"'d'"'"'' '-D_PICOLIBC_SCANF='"'"'d'"'"'' -Wl,--gc-sections -nostdlib -T /local/mnt/workspace/garvgupt/v2_cpullvm/qualcomm-software/build/multilib-builds/picolibc/riscv32imac_ilp32/build/picolibc.ld -lclang_rt.builtins -Wl,--end-group`

jonathonpenix:
IIUC, picolibc's docs seem fairly clear about the [expected ordering](https://github.com/picolibc/picolibc/blob/main/doc/using.md#semihosting) so I'd rather we lean towards being consistent between targets even if it currently happens to work for Arm/AArch64 for whatever compiler-rt happens to pull out of libc/libsemihost for those targets/tests.

Besides that, I guess it feels weird to me that we're going to have an extra `-lc` peppered in, but I guess it seems like it should be harmless? Especially given that it'll just go into a `--start-group`, etc. for libc++. But, not sure how else to work around this offhand, so I'm fine with this.

Edit: one other drive-by thought is that we might be able to limit this to compiler-rt only rather than having it percolate through to libc++ and picolibc too.

quic-garvgupt:
>consistent between targets even if it currently happens to work for Arm/AArch64

to be more clear, I should remove riscv specific target check?

>I guess it feels weird to me that we're going to have an extra `-lc` peppered in

The expected ordering link  also have -lc before -lsemihost on the comamnd line. Can you pls elaborate why it is weird?

>one other drive-by thought is that we might be able to limit this to compiler-rt only rather than having it percolate through to libc++ and picolibc too

I didnt understood this comment. Can you pls elaborate? This change for now affects only compiler-rt  tests and maybe libc++ tests as well but definitely not picolibc tests

quic-garvgupt:
PFB the complete chain on how this var gets used and expanded to full link line for compiler-rt tests

**qualcomm-software/embedded-runtimes/CMakeLists.txt** (CPULLVM only file)
`line 176-182:`
set(semihost_link_flags "-lc -lsemihost")          # RISC-V
set(test_link_flags "-fuse-ld=eld -nostartfiles -l${picocrt} ${semihost_link_flags} -T picolibcpp.ld")

So test_link_flags becomes ... -lcrt0-semihost -lc -lsemihost -T picolibcpp.ld (the -lc is inserted between -lcrt0-semihost and -lsemihost).

`Line 321`: compiler_rt_test_flags = "${lib_compile_flags} ${test_link_flags}"
`Line 327`: -D**COMPILER_RT_TEST_COMPILER_CFLAGS**=${compiler_rt_test_flags} passed as cmake cache var to the compiler-rt ExternalProject.

**compiler-rt/cmake/config-ix.cmake**
`Line 315`: macro(get_test_cc_for_arch arch cc_out cflags_out)
    if(... ${arch} MATCHES "arm|aarch64|riscv32|riscv64")
    set(${cc_out}     ${COMPILER_RT_TEST_COMPILER})
    set(${cflags_out} ${**COMPILER_RT_TEST_COMPILER_CFLAGS**})  <-- here
    
The above macro is called from the below file as:
**compiler-rt/test/builtins/CMakeLists.txt**
`Line 54:` 
get_test_cc_for_arch(${arch} BUILTINS_TEST_TARGET_CC
                                    BUILTINS_TEST_TARGET_CFLAGS)
     **BUILTINS_TEST_TARGET_CFLAGS** = COMPILER_RT_TEST_COMPILER_CFLAGS
     
So `BUILTINS_TEST_TARGET_CFLAGS` receives the value and then in test/builtins/Unit/lit.site.cfg.py.in:5
config.target_cflags = "@BUILTINS_TEST_TARGET_CFLAGS@"

test/builtins/Unit/lit.cfg.py:151:  reads `config.target_cflags` into `target_cflags`, builds the %clang_builtins substitution (which carries the test_link_flags), and defines %librt.
Line 110: 
config.substitutions.append(
            ("%librt ", "-lm -Wl,--start-group " + base_lib + " -lc -Wl,--end-group ")
        )

Therefore final link line assembles something like this:

clang --target=riscv32-unknown-elf -march=rv32imac -mabi=ilp32 -fPIC --sysroot .../tmp_install \
       -ffunction-sections -fdata-sections -ftls-model=local-exec \
       ┌─── from test_link_flags (the patch) ─────────────────────────────────┐
       -fuse-ld=eld -nostartfiles  -lcrt0-semihost  -lc  -lsemihost  -T picolibcpp.ld
       └────────────────────────────────────────────────────────┘
       -Wthread-safety ... -fforce-enable-int128 -DCOMPILER_RT_HAS_FLOAT16 \
       -nobuiltininc -I.../compiler-rt/include ... -fno-builtin -I .../lib/builtins -nodefaultlibs \
       udivsi3_test.c \
       ┌─── from %librt (lit.cfg.py) ─────────────────────────────────────────┐
       -lm  -Wl,--start-group  .../libclang_rt.builtins.a  -lc  -Wl,--end-group
       └──────────────────────────────────────────────────────────┘
       -o .....
       
       
Hopefully this demonstrates how the link command is expanded for compiler-rt tests and the relative position of libc, builtins, and other semihosting libraries wrt each other.

jonathonpenix:
> > I guess it feels weird to me that we're going to have an extra `-lc` peppered in
> 
> The expected ordering link also have -lc before -lsemihost on the comamnd line. Can you pls elaborate why it is weird?

"Weird" was maybe the wrong word, but this is my thinking:

For libc++, my guess is that actual linker invocation is going to have something like this with this change:
```
ld.eld ... --start-group -lc -lm -lclang_rt.builtins -lc -lsemihost --endgroup`
```
Which, I guess the `-lc` repetition just seems redundant (and the ordering was already correct/has everything in a group before this change). But, this should just be cosmetic/harmless AFAICT.

For compiler-rt, based on what you wrote above, my understanding is that the linker invocation is going to look something like this:
```
ld.eld ... -lcrt0-semihost -lc -lsemihost -T picolibcpp.ld udivsi3_test.o -lm -Wl,--start-group .../libclang_rt.builtins.a -lc -Wl,--end-group
```
To me, it seems like the correct ordering should really just be `-lsemihost` moved after the `-lc` at the end and no extra `-lc` added. The (more theoretical at this point) concern being that, with the ordering above, the `-lc -lsemihost` only catches things required by the `-lcrt0-semihost`. If ex: udivsi3_test.o (which I'm assuming is the actual test and/or bulk of the test executable?) pulls something out of libc that depends on libsemihost (and wasn't also required by crt0-semihost), we should hit errors again.

But, I understand your comments above about not being able to achieve this link order without touching upstream code + this works for the tests as they exist today. We can always circle back to this when we have more time

>> one other drive-by thought is that we might be able to limit this to compiler-rt only rather than having it percolate through to libc++ and picolibc too
>
> I didnt understood this comment. Can you pls elaborate? This change for now affects only compiler-rt tests and maybe libc++ tests as well but definitely not picolibc tests

You're right, I should have double checked on picolibc.

I just meant that this could also be hacked in via ex: `compiler_rt_test_flags` directly to limit it to compiler-rt only. But again, the impact on libc++ should be harmless (which, I _think_ it should make it through, but I might be wrong) so it seems fine to me.

jonathonpenix (approved):
LGTM, thanks @quic-garvgupt

I guess longer term maybe we can look into whether there's some way to support this better upstream (similar to libc++? or via multilib or config file or something?) but I'm happy with as-is for now if the tests are passing

lenary (approved):
LGTM.

Thanks for the explanation and clarity, I'm happy for us to proceed with this for 22.

I hope we come back to this with thoughts of how to solve the placement of `-lsemihost` in both the driver and the tests in the future.

quic-garvgupt:
@apazos can you pls merge this PR on my behalf?

quic-garvgupt:
>We can always circle back to this when we have more time
I hope we come back to this with thoughts of how to solve the placement of -lsemihost in both the driver and the tests in the future.

I will create a ticket to track this for the next release

>I just meant that this could also be hacked in via ex: compiler_rt_test_flags directly to limit it to compiler-rt only. 

Thanks for clarifying. I thought about that too but was a bit 50-50 there because link flags are assembled in `test_link_flags` and was not much aware of its effect back then on libc++.

apazos (approved):

PR #407, opened 2026-06-10
[cpullvm][Github] Don't build musl overlay for PR premerge testing

jonathonpenix:
This is one easy option to reduce the burden of PRs on our infrastructure--rather than each premerge build using two builders, removing the musl-embedded overlay should get us down to one.

Given that we're discussing removing the musl-embedded overlay from our releases anyway, it might be a worthwhile tradeoff (we'd still be building this in each of our nightlies).

### DIFF
--- a/.github/workflows/linux-premerge.yml
+++ b/.github/workflows/linux-premerge.yml
@@ -39,11 +39,6 @@ jobs:
             target_os: Linux-x86_64
             runner: cpullvm-ubuntu24-x86_64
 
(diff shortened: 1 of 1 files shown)


Yethendra (approved):
Looks good to me

pranav4330 (approved):
Got it, This will help thanks.

apazos (approved):

jonathonpenix:
Everyone seems to be ok with this, so will merge

PR #412, opened 2026-06-12
[cpullvm][Multilib] Fix armv7a and armv8a multilib matching

jonathonpenix:
There's three intertwined things here:

1. Our armv7a/armv8a variants required a normalized `thumbv*` triple to match the variant, even though there isn't an issue mixing arm and thumb modes for the purposes of our libraries.

   Fix this by switching to requiring `armv*` triples as the rest of our multilib setup is written canonicalizing towards `armv*` so generally already handles the mappings we want/need.
2. Add `thumbv8a` -> `armv8a` triple mappings. We didn't have these previously as we inherited our setup from ATfE. I think they omit(ted) them only as they don't have any armv8a variants [1].
3. With 1 and 2 sorted, we have to move our armv8_soft_neon variant for it to layer correctly with our armv7a_soft_neon variant.

[1] https://github.com/arm/arm-toolchain/commit/29fd17cf82ce40e57f7d4d96a2b36b2048eecffe

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -36,19 +36,19 @@
             "flags": "--target=aarch64-unknown-none-elf -march=armv8.5-a -march=armvX+nofp -march=armvX+nosimd -mbranch-protection=pac-ret+leaf+b-key+bti -mabi=aapcs-soft"
         },
         {
-            "variant": "armv8_soft_neon",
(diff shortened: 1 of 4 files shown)


jonathonpenix:
This is https://github.com/qualcomm/cpullvm-toolchain/pull/410/ ported to 22.x

jonathonpenix:
#410 has been merged and it sounds like this is working as expected in our release build.

apazos (approved):

PR #416, opened 2026-06-15
[cpullvm] Use internal lit shell

jonathonpenix:
Using an external shell is now deprecated in LLVM, see:
- https://github.com/llvm/llvm-project/commit/57109befac92811d2253109242ca6fa69c961fb2
- https://discourse.llvm.org/t/rfc-removal-of-the-lit-external-shell/90951
- https://discourse.llvm.org/t/rfc-enabling-the-lit-internal-shell-by-default/80179

This is breaking our builds as we explicitly force the external shell in a few places.

We can either:
- Bypass the errors and continue using the external shell via `force_execute_external` (until LLVM 24).
- Remove our dependencies on the external shell and migrate to the internal one.

I'm not entirely sure why we have the external shell dependency currently, but from what I can see we don't have anything that should be incompatible with lit's internal shell. So try that fix as it is what we need to do long term anyway.

### DIFF
--- a/qualcomm-software/embedded-runtimes/test-support/gen-picolibc-lit-tests.py
+++ b/qualcomm-software/embedded-runtimes/test-support/gen-picolibc-lit-tests.py
@@ -90,7 +90,7 @@ lit.llvm.initialize(lit_config, config)
 
 config.name = "%CONFIG_NAME%"
 config.suffixes = [".test"]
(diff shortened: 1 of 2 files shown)


apazos:
For now this solution aligns with the solution discussed in Arm embedded toolchain, so it looks fine. Let's keep monitoring that discussion and update accordingly.

apazos (approved):

PR #438, opened 2026-06-22
[cpullvm] mute "Up-to-date" messages from cmake

bojle:
The CMAKE_INSTALL_MESSAGE variable disables printing verbose,

"[Up-to-date] Lib... "

which otherwise pollute logs. This variable is set to NEVER in the top level cmake but never passed down to the ExternalProjects' cmake file.

This patch fixes that.


bojle:
@jonathonpenix

jonathonpenix (approved):
LGTM, thanks!

In case it is helpful for others, this takes our log from ~580k lines to ~360k lines, so it does make a substantial difference.

We might want to make similar changes for our Linux runtimes eventually, but that is another can of worms/I'm happy with dealing with that later.

PR #449, opened 2026-06-25
[cpullvm] Split `ninja check-*` lines in test scripts.

jonathonpenix:
`check-all-llvm-toolchain` runs *all* enabled tests, which we don't want while we're trying to ex: enable libcxxabi tests before libcxx tests. And, the libcxx tests should be fairly slow to run so we won't always want those.

So, split `check-all-llvm-toolchain` into `check-all` and `check-llvm-toolchain`.

For the tests we have enabled today, this should be NFC:
* check-all covers all LLVM-ish projects + eld
* check-llvm-toolchain covers picolibc + compiler-rt + multilib tests

So, everything we have enabled is covered.

Note that this isn't strictly necessary for test.ps1 at the moment since QEMU testing is disabled on Windows. But, I'm splitting this anyway just to make it clearer what our intention is if we ever get to QEMU on Windows.


jonathonpenix:
@bojle if this goes in I think you'll have to add `ninja check-cxxabi` to `test.sh` as part of #341, just a heads up

jonathonpenix:
Failures unrelated: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/28143261759/job/83344877627

apazos (approved):

bojle (approved):
LGTM

apazos (approved):

PR #452, opened 2026-06-25
[cpullvm] Temporarily disable localization for our picolibc builds

jonathonpenix:
This is broken after https://github.com/llvm/llvm-project/commit/8a531c3608c722ad529be448d6ecef06ba107228

Temporarily disable localization until this is resolved upstream. I think this might break some Zephyr testing that uses C++, but will hopefully unblock our builders in the meantime.


jonathonpenix:
It looks like there's a proposed fix for this upstream already in PR 205882, not sure how soon this will be merged/if we need our builders back immediately

PR #456, opened 2026-06-29
[cpullvm] Add workflow step to log repository commit SHAs

sriyalamar:
Add a new step in CI workflows to capture and print the commit SHA for all repositories involved in the build.

This improves traceability by allowing us to track the exact combination of repository revisions used in each nightly or CI build.

The logged information can be used for debugging, reproducibility, and auditing build configurations.

Change-Id: Ida520cc94eb81c3fbde14d6d769e27dd8c81ac57

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -50,6 +50,10 @@ jobs:
       - name: Build
         run: ./qualcomm-software/scripts/build.sh
 
+      - name:  echo git repo versions
+        if: always()
+        run: cat build/VERSION.txt
+
       - name: Upload built packages
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         if: success()
--- a/.github/workflows/linux-premerge.yml
+++ b/.github/workflows/linux-premerge.yml
@@ -55,6 +55,10 @@ jobs:
       - name: Build toolchain
         run: ./qualcomm-software/scripts/${{ matrix.build_script }}
 
+      - name:  echo git repo versions
+        if: always()
+        run: cat build/VERSION.txt
+
       - name: Upload built packages
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         if: success()


.github/workflows/nightly.yml
```diff
@@ -59,6 +59,12 @@ jobs:
           if-no-files-found: error
           retention-days: 2
 
+      - name: Collect repo versions
+        if: always()
+        run: |
+          ./qualcomm-software/scripts/git-repo-versions.sh > repo_versions.txt
```
navaneethshan: Why do you have to put it into repo_version.txt and then print it, directly call the sh script which will print the same content onto the console?

navaneethshan:
Aren't we already capturing it in VERSION.txt as below?

nshanmug@hu-nshanmug-lv:/prj/llvm-arm/hexbuild_home/nightly/install/cpullvm/qualcomm-software/x86_linux/latest$ cat VERSION.txt 
CPULLVM Toolchain 

Sources:
* cpullvm: https://github.com/qualcomm/cpullvm-toolchain (commit 289a2f5462fff89ac9ab2b237fde668d249c508c)
* eld: https://github.com/qualcomm/eld (commit db955252159d25109849c5c5d0efa778adf21504)
* picolibc: https://github.com/picolibc/picolibc.git (commit 51a8b32857e75345c37652a80b5cda98b28d69e5)
* musl: https://git.musl-libc.org/git/musl (commit 0784374d561435f7c787a555aeab8ede699ed298)
* musl-embedded: https://github.com/qualcomm/musl-embedded (commit 9eafd9adeff02200cecb6a8a745898e33301f4f6)

jonathonpenix:
Two drive-by comments:
1. Can you expand a bit on what this is intended to be used for? If the build succeeds, this info is available in the build already like Navaneeth mentioned. <br><br>If the build fails (so we don't upload it) is VERSION.txt actually guaranteed to be generated? I'm not entirely sure how that works offhand.<br><br>FWIW though, I think this information is already available in the log as part of cloning the repos, so if it is purely for debugging these cases that info already exists (though it is a bit buried so I don't mind if we make it more visible)
2. If we make a change like this, please make it across all the nightly configurations (the native runtime one is missing right now)

jonathonpenix:
Another misc. thought: there's some discussion around this elsewhere, but if we proceed with switching eld to a specific SHA and updating versions.json directly, I think versions.json is a pretty accurate source of truth (all of the repos would be pointing at a particular SHA or tag) so this might be kind of redundant in that sense too.

(picolibc will also be switched to a specific sha soon-ish I think, and I'm not sure if we have plans to point to mainline any time soon)

PR #468, opened 2026-07-03
[cpullvm] Update Linux kernel source location to stable git repository

Yethendra:


efriedma-quic (approved):
LGTM

jonathonpenix (approved):
LGTM to unblock our builders.

I'm assuming at some point we'd want to switch back to a tarball though (either the old link, or the same scm thing ships tarballs I think) as I assume it is just less to download?

Also, I think at one point @efriedma-quic had a patch that would also cache this stuff: https://github.com/qualcomm/cpullvm-toolchain/pull/247

We might want to revisit that--I think that would help a bit (IIUC we'd only hit the kernel servers if something happened to our cache?) but I might be misunderstanding

pranav4330 (approved):
Looks good. I had made similar changes in my PR, but for some reason they didn't go through. Another solution I was looking was to keep the tar file in a repository folder, so it remains accessible whenever we need it. but this will work now.

apazos:
Thanks for quickly pushing a fix for the unavailability of the site

PR #472, opened 2026-07-06
[cpullvm][Github] Update upload-artifact action to v7.0.1

jonathonpenix:
Updating to the latest release. The primary motivation for doing this is it adds a `archive` option that let's us skip the extra zip step, which should simplify our workflows slightly.


jonathonpenix:
The main thing I wanted this for (#473) doesn't actually seem like a good idea. So closing for now

PR #473, opened 2026-07-06
[cpullvm][Github] Don't re-zip our uploaded toolchains

jonathonpenix:
Right our toolchains are packaged as a zip or tarfile which is then
*re-zipped* by Github.

So we have `cpullvm-packages-*` as the uploaded artifact that unzips to
a `build/` folder with the real toolchain package inside.

Github's upload-artifact v7.0.1 added a new `archive` option that lets us
avoid this situation. This should make it a bit clearer what we're actually
intending to upload and avoids some extra steps when trying to extract the
toolchain after downloading.

=====

Stacked on https://github.com/qualcomm/cpullvm-toolchain/pull/472

### DIFF
--- a/.github/workflows/linux-premerge.yml
+++ b/.github/workflows/linux-premerge.yml
@@ -59,9 +59,9 @@ jobs:
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         if: success()
         with:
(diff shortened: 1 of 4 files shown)


jonathonpenix:
Going to close this for now. Our musl embedded overlays rely on being able to upload multiple things at once. I think we're going to see the same thing for picolibc version overlays. And we've talked about other dev tool overlay-ish things in the past, so I'm not sure this is actually a good idea long term.

PR #477, opened 2026-07-07
[embedded-runtimes] Remove SHARED_OUTPUT_NAME

lenary:
Upstream commit 885ef9daac3570ddfdd20cc82296eeae5b2c6613 should mean that we don't need to set these any more, as the shared outputs should be suppressed on our targets.


jonathonpenix (approved):
Nice, thanks!

PR #478, opened 2026-07-07
[cpullvm] Add versions.json repository update script

sriyalamar:
Added a standalone script to update repository revisions stored in
versions.json.

Also updated the automerge workflow to run the dependency version update
step after successfully merging upstream LLVM commits.

This helps track dependency revision history and simplifies bisection of
build and nightly test failures.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -43,6 +43,17 @@ jobs:
           git fetch origin ${{ matrix.branches.from_branch }}:${{ matrix.branches.from_branch }}
           git pull --unshallow --ff-only origin ${{ matrix.branches.to_branch }}
       - name: Run automerge
-        run: python3 qualcomm-software/ci/automerge.py --project-name ${{ github.repository }} --from-branch ${{ matrix.branches.from_branch }} --to-branch ${{ matrix.branches.to_branch }} --verbose
+        run: |
+          python3 qualcomm-software/ci/automerge.py \
+            --project-name ${{ github.repository }} \
+            --from-branch ${{ matrix.branches.from_branch }} \
+            --to-branch ${{ matrix.branches.to_branch }} \
+            --verbose
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+      - name: Update dependency repo versions
+        run: |
+          python3 qualcomm-software/ci/update_repo_versions.py \
+            --to-branch ${{ matrix.branches.to_branch }}
         env:
           GH_TOKEN: ${{ steps.generate-token.outputs.token }}
--- a/qualcomm-software/ci/update_repo_versions.py
+++ b/qualcomm-software/ci/update_repo_versions.py
@@ -1,0 +1,194 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import argparse
+import json
+import logging
+import subprocess
+from pathlib import Path
+
+logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
+logger = logging.getLogger(__name__)
+
+REMOTE_NAME = "origin"
+
+
+def run_cmd(cmd: list[str], cwd: Path | None = None) -> str:
+    result = subprocess.run(
+        cmd,
+        check=True,
+        capture_output=True,
+        text=True,
+        cwd=cwd,
+    )
+    return result.stdout.strip()
+
+
+def update_repo_version(
+    repo_path: Path,
+    repo_name: str,
+    repo_info: dict,
+) -> None:
+    versions_file = repo_path / "qualcomm-software" / "versions.json"
+
+    tag_type = repo_info.get("tagType")
+    repo_url = repo_info.get("url")
+    current_tag = repo_info.get("tag")
+
+    logger.info("Repo: %s", repo_name)
+    logger.info("URL: %s", repo_url)
+    logger.info("tagType: %s", tag_type)
+    logger.info("Current tag: %s", current_tag)
+
+    # Only update repositories already using commit hashes
+    if tag_type != "commithash":
+        logger.info(
+            "Skipping %s because tagType is '%s'",
+            repo_name,
+            tag_type,
+        )
+        return
+
+    tracking_branch = repo_info.get("trackingBranch")
+
+    if not tracking_branch:
+        logger.info(
+            "Skipping %s because 'trackingBranch' is not set",
+            repo_name,
+        )
+        return
+
+    branch_ref = f"refs/heads/{tracking_branch}"
+    logger.info("Tracking ref: %s", branch_ref)
+
+    latest_sha = run_cmd(
+        ["git", "ls-remote", repo_url, branch_ref]
+    ).split()[0]
+
+    logger.info("Latest SHA: %s", latest_sha)
+
+    if latest_sha == current_tag:
+        logger.info("%s already up to date", repo_name)
+        return
+
+    # Reload the file to reflect any updates from previous repo iterations
+    with open(versions_file, encoding="utf-8") as f:
+        data = json.load(f)
+
+    # Update only tag
+    data["repos"][repo_name]["tag"] = latest_sha
+
+    with open(versions_file, "w", encoding="utf-8") as f:
+        json.dump(data, f, indent=2)
+        f.write("\n")
+
+    logger.info("Updated versions.json for %s", repo_name)
+
+    # Stage and commit one commit per repo
+    run_cmd(
+        ["git", "-C", str(repo_path), "add", str(versions_file)]
+    )
+
+    commit_msg = f"Automerge: Update {repo_name} to {latest_sha}"
+
+    run_cmd(
+        [
+            "git",
+            "-C",
+            str(repo_path),
+            "commit",
+            "-m",
+            commit_msg,
+        ]
+    )
+
+    logger.info("Created commit: %s", commit_msg)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "--repo-path",
+        default=Path.cwd(),
+        type=Path,
+    )
+
+    parser.add_argument(
+        "--to-branch",
+        required=True,
+        help="Target cpullvm branch",
+    )
+
+    parser.add_argument(
+        "--dry-run",
+        action="store_true",
+    )
+
+    args = parser.parse_args()
+
+    versions_file = args.repo_path / "qualcomm-software" / "versions.json"
+
+    logger.info("Switching to branch %s", args.to_branch)
+    run_cmd(["git", "-C", str(args.repo_path), "switch", args.to_branch])
+
+    logger.info("Loading %s", versions_file)
+
+    with open(versions_file, encoding="utf-8") as f:
+        data = json.load(f)
+
+    repos = data.get("repos", {})
+
+    original_head = run_cmd(["git", "-C", str(args.repo_path), "rev-parse", "HEAD"])
+
+    for repo_name, repo_info in repos.items():
+        try:
+            update_repo_version(
+                repo_path=args.repo_path,
+                repo_name=repo_name,
+                repo_info=repo_info,
+            )
+        except subprocess.CalledProcessError as error:
+            logger.error(
+                'Failed to run command: "%s"\nstdout:\n%s\nstderr:\n%s',
+                " ".join(error.cmd),
+                error.stdout,
+                error.stderr,
+            )
+            raise SystemExit(1)
+
+        except Exception as error:
+            logger.exception(
+                "Failed to update repository '%s': %s",
+                repo_name,
+                str(error),
+            )
+            raise SystemExit(1)
+
+    current_head = run_cmd(["git", "-C", str(args.repo_path), "rev-parse", "HEAD"])
+
+    if current_head == original_head:
+        logger.info("All tracked repositories are already up to date")
+        return
+
+    if args.dry_run:
+        logger.info("Dry run; skipping push")
+        return
+
+    run_cmd(
+        [
+            "git",
+            "-C",
+            str(args.repo_path),
+            "push",
+            REMOTE_NAME,
+            f"HEAD:{args.to_branch}",
+        ]
+    )
+    logger.info("Push complete")
+
+
+if __name__ == "__main__":
+    main()
--- a/qualcomm-software/versions.json
+++ b/qualcomm-software/versions.json
@@ -22,8 +22,9 @@
     },
     "eld": {
       "url": "https://github.com/qualcomm/eld",
-      "tagType": "branch",
-      "tag": "main"
+      "tagType": "commithash",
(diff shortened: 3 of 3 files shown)


sriyalamar:
Hi All, As part of the verification in my fork repository, I have confirmed that the Git log history for qualcomm-software/versions.json is visible and traceable.
I also verified that checking out a specific cpullvm-toolchain commit correctly corresponds to the respective ELD SHA checkout.
Please let me know if any additional details are required or if there are other verification scenarios that I should include as part of this validation.
Thank you.

lenary:
Please remove the commit message hook from your local repo, and remove the `Change-Id:` line from your commit messages and description. To remove the hook, you can `rm .git/hooks/commit-msg` (or chmod it to make it not executable)

qualcomm-software/ci/automerge.py
```diff
@@ -208,6 +208,68 @@ def is_worktree_clean(git_repo: Git) -> bool:
...
+        logger.info("Latest eld SHA: %s", eld_sha)
+
+        versions_file = git_repo.get_repo_path() / "qualcomm-software" / "versions.json"
+        logger.info("Versions file path: %s", versions_file)
+
+        # Load JSON
+        with open(versions_file) as f:
+            data = json.load(f)
+
+        current_sha = data["repos"]["eld"].get("tag", "")
```
jonathonpenix: Can you also check that the `tagType` is `commithash` before we do any updating?

If for some reason we switch it to a branch, tag, etc. I don't think we should be overwriting it.

qualcomm-software/ci/automerge.py
```diff
@@ -208,6 +208,68 @@ def is_worktree_clean(git_repo: Git) -> bool:
...
+        logger.info("Current eld SHA in file: %s", current_sha)
+
+        if current_sha == eld_sha:
+            logger.info("Eld SHA already up to date. Skipping update.")
+            return
+
+        # Update values
+        data["repos"]["eld"]["tagType"] = "commithash"
```
jonathonpenix: Similar to above, I don't think we should be overwriting the `tagType`, only the sha _if_ `tagType` was already `commithash`

qualcomm-software/ci/automerge.py
```diff
@@ -208,6 +208,68 @@ def is_worktree_clean(git_repo: Git) -> bool:
     status_output = git_repo.run_cmd(["status", "--porcelain"]).strip()
     return len(status_output) == 0
 
+def update_eld_version(git_repo: Git, to_branch: str, dry_run: bool) -> None:
```
jonathonpenix: I think it's worth defining this is in a more generic way (ie it should work for any of the repos we have in versions.json) then we can just call the function with the repo-specific information. (So pass in ex: "eld" to this function and we can look up everything else from versions.json--the url, etc. is already there).

Then we can use it for other repos pretty easily.

qualcomm-software/ci/automerge.py
```diff
@@ -208,6 +208,68 @@ def is_worktree_clean(git_repo: Git) -> bool:
...
+    try:
+        logger.info("Updating eld sha in versions.json")
+
+        # ensure we are on target branch
+        git_repo.run_cmd(["switch", to_branch])
+
+        # Get latest SHA
+        result = subprocess.run(
+            ["git", "ls-remote", "https://github.com/qualcomm/eld", "HEAD"],
```
jonathonpenix: The repo is available in versions.json--we we only need to know the repo "name"

qualcomm-software/ci/automerge.py
```diff
@@ -281,6 +343,8 @@ def main():
                 args.dry_run,
                 args.verbose,
             )
+        # After all merges, update eld revision
```
jonathonpenix: It _might_ make sense to just implement this version stuff in a separate python script since it is fairly unrelated to the rest of automerge.py

Then add it as a separate step after automerge.py in automerge.yml

(if there's utilities you want you can always import them I guess)

Will leave up to you though if you want to do this

jonathonpenix (commented):
Think there's a few things we need to be careful about:
- If we go forward with something like this, it probably needs to work on CPULLVM mainline and release branches. So right now we are looking at the HEAD of eld (main) which isn't what we want for the branch. (Thanks Sam for mentioning this)
- Whatever we decide, I think it is worth doing this in such a way that it works for any/all of the repos in our versions.json. It should be very little extra to make that work I think (just passing in a name parameter to `update_eld_version()` rather than hardcoding "eld" throughout).

I think something like this makes sense to me (which, again some of this is stolen from talking with Sam so thanks Sam on that too):
* "introduce" a new entry in versions.json that a) indicates that that repo should be auto-updated and b) indicates the branch that it should be tracking<br><br>So maybe something like "autoTrackBranch": "mybranch"
*  In our script (automerge.py currently, or the new script if you split it out), iterate over the repos in versions.json.<br><br>for each repo, if "autoTrackingBranch" is present, then call a generic version of `update_eld_version()` with the repo name

I think this ends up being fairly robust in terms of handling different repos with different branches across different cpullvm release branches?

Would be curious to hear if others have thoughts on this

qualcomm-software/ci/automerge.py
```diff
@@ -208,6 +208,68 @@ def is_worktree_clean(git_repo: Git) -> bool:
...
+        git_repo.run_cmd(["add", str(versions_file)])
+        # Check if there are staged changes
+        proc = subprocess.run(
+            ["git", "-C", str(git_repo.get_repo_path()), "diff", "--cached", "--quiet"]
+        )
+        if proc.returncode == 0:
+            logger.info("No changes to versions.json")
+            return
+        logger.info("Changes detected, committing...")
+        git_repo.run_cmd(["commit", "-m", f"Update eld to {eld_sha}"])
```
jonathonpenix: Is the final commit message going to include something like how we have 'Automerge:' in our normal auto-merged merge commits?

I think it'd be helpful to a have something indicating it is coming from an automated workflow.

sriyalamar:
Thank you @jonathonpenix  and @lenary for the review and suggestions.

I've addressed the review comments and updated the implementation. Verification is currently in progress.
Could you please confirm whether the updated flow looks good? If so, I'll follow up with the verification details once testing is complete.

https://github.com/qualcomm/cpullvm-toolchain/pull/478/changes/fdfc5cf38c0a5c1b1a1b76cb7096f88b95ab2db4

.github/workflows/automerge.yml
```diff
@@ -44,3 +44,5 @@ jobs:
         run: python3 qualcomm-software/ci/automerge.py --project-name ${{ github.repository }} --from-branch ${{ matrix.branches.from_branch }} --to-branch ${{ matrix.branches.to_branch }} --verbose
         env:
           GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+      - name: Update eld version
+        run: python3 qualcomm-software/ci/update_versions.py --repo-name eld
```
jonathonpenix: I think you still need something like the `--to-branch` since this workflow will run on cpullvm's main branch (not in the release branches), but it needs to possibly make changes on the cpullvm release branches.

To make this work for branches, we'd add the relevant cpullvm from/to branches to the matrix above.

(I think you were checking out the `to-branch` earlier when it was in automerge.py, but it looks like it was dropped in the last update)

qualcomm-software/ci/update_repo_versions.py
```diff
@@ -0,0 +1,182 @@
...
+            "Skipping %s because tagType is '%s'",
+            repo_name,
+            tag_type,
+        )
+        return
+
+    logger.info("Fetching latest commit SHA")
+
+    latest_sha = run_cmd(
+        ["git", "ls-remote", repo_url, "HEAD"]
```
jonathonpenix: HEAD isn't always the right thing--it might need to be a different branch, etc.

I think the easiest way around this right now is just to add to `versions.json` something like a `trackingBranch` or `autoTrackingBranch`--see the comments earlier on this

qualcomm-software/ci/automerge.py
```diff
@@ -208,7 +208,6 @@ def is_worktree_clean(git_repo: Git) -> bool:
     status_output = git_repo.run_cmd(["status", "--porcelain"]).strip()
     return len(status_output) == 0
 
-
```
jonathonpenix, line 211: Extra whitespace change, can be removed

qualcomm-software/ci/update_repo_versions.py
```diff
@@ -0,0 +1,182 @@
...
+    ).split()[0]
+
+    logger.info("Latest SHA: %s", latest_sha)
+
+    if latest_sha == current_tag:
+        logger.info("%s already up to date", repo_name)
+        return
+
```
jonathonpenix, line 75: Extra newline, can be removed I think

qualcomm-software/ci/update_repo_versions.py
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python3
```
jonathonpenix, line 1: Add license lines

.github/workflows/automerge.yml
jonathonpenix: One additional comment, if we do the `trackingBranch` thing, I think we can drop the `--repo-name` flag and just let this be controlled in one spot in `versions.json`

.github/workflows/automerge.yml
```diff
@@ -41,6 +41,18 @@ jobs:
...
+            --project-name ${{ github.repository }} \
+            --from-branch ${{ matrix.branches.from_branch }} \
+            --to-branch ${{ matrix.branches.to_branch }} \
+            --verbose
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+      - name: Update eld version
+        run: |
+          python3 qualcomm-software/ci/update_repo_versions.py \
+          --repo-name eld \
```
jonathonpenix: Please see this comment: https://github.com/qualcomm/cpullvm-toolchain/pull/478/#discussion_r3555747381

I don't think we should be passing in repo names at this level--just let it be controlled via the presence of the `trackingBranch` in versions.json.

Just have update_repo_versions.py iterate over every entry in versions.json and only update those entries/repos that have `commithash` + `trackingBranch` present.

This has the benefits of letting `versions.json` be the single source of truth and this should "naturally" handle if we want different behaviors across different cpullvm branches.

qualcomm-software/ci/update_repo_versions.py
jonathonpenix: `commithash` without `trackingBranch` should be fine, it just means we should just skip updating anything.

So I don't think this should be an error.

.github/workflows/automerge.yml
```diff
@@ -41,6 +41,18 @@ jobs:
...
-        run: python3 qualcomm-software/ci/automerge.py --project-name ${{ github.repository }} --from-branch ${{ matrix.branches.from_branch }} --to-branch ${{ matrix.branches.to_branch }} --verbose
+        run: |
+          python3 qualcomm-software/ci/automerge.py \
+            --project-name ${{ github.repository }} \
+            --from-branch ${{ matrix.branches.from_branch }} \
+            --to-branch ${{ matrix.branches.to_branch }} \
+            --verbose
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+      - name: Update eld version
```
jonathonpenix: Please also update this to be not eld-specific

qualcomm-software/versions.json
jonathonpenix: Sorry, I said to add the `trackingBranch` earlier but we should either:
* Add `trackingBranch` _and_ switch the tag type/tag so that once this is merged it should start working/running for eld -OR-
* Omit `trackingBranch` and switch over to `trackingBranch` + `commithash` in a separate PR.

Either way is fine with me--will leave it up to you.

But I think we should go one way or another instead of leaving it in the middle.

jonathonpenix:
Please also remove the change-id line from the PR description

qualcomm-software/ci/update_repo_versions.py
jonathonpenix: One commit per repo please, the push can happen at the end still

If something breaks after a repo is updated, being able to ex: bisect back to a single commit that is modifying only one repo should make life easier for folks.

.github/workflows/automerge.yml
```diff
@@ -41,6 +41,17 @@ jobs:
...
+            --project-name ${{ github.repository }} \
+            --from-branch ${{ matrix.branches.from_branch }} \
+            --to-branch ${{ matrix.branches.to_branch }} \
+            --verbose
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+      - name: Update dependency repo versions
+        run: |
+          python3 qualcomm-software/ci/update_repo_versions.py \
+          --to-branch ${{ matrix.branches.to_branch }}
```
jonathonpenix: Please indent this similar to what you did above if that's the preferred formatting

qualcomm-software/ci/update_repo_versions.py
jonathonpenix: CPULLVM is under the Apache license that was here earlier, please ignore the license checker failures--it assumes a default license of BSD 3 which isn't correct here.

jonathonpenix:
The parts updating SHAs in versions.json look good to me

It looks like this is currently/was just recently bundled with another change in the nightlies though? Not sure if that is for testing something or accidental

I'm assuming that should be removed before merging.

jonathonpenix (approved):
LGTM, thanks!

I'm assuming a rebase will remove the clang test failure since our nightlies are clean.

(And FWIW, if you want to update the eld sha just before merging that is fine with me and I don't think you need to wait for reapproval. Otherwise I'm not too worried about using an older eld sha for a few minutes before it auto-updates)

sriyalamar:
Thank you so much, @jonathonpenix , for the review and suggestions.
I have updated the code to include the ELD SHA update for the 23.x branch as well.
I have created a separate PR for the 23.x branch to support this change: https://github.com/qualcomm/cpullvm-toolchain/pull/513 
Please check it once.

.github/workflows/automerge.yml
```diff
@@ -43,6 +43,18 @@ jobs:
...
+          python3 qualcomm-software/ci/automerge.py \
+            --project-name ${{ github.repository }} \
+            --from-branch ${{ matrix.branches.from_branch }} \
+            --to-branch ${{ matrix.branches.to_branch }} \
+            --verbose
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+      - name: Update dependency repo versions
+        run: |
+          git switch ${{ matrix.branches.to_branch }}
```
jonathonpenix: Why is this needed?

The checkout above is checking out the `to_branch` so we should already be on this branch, no?

And we:
* Explicitly switch branches again in the `update_repo_version.py`
* Are explicit about what branch we're pushing to in `update_repo_version.py`

In `automerge.py` I also see that there's an explicit switch in the script, but no `git switch` in the workflow step for that so again seems surprising that we'd have to add it here.

jonathonpenix (commented):
>I have updated the code to include the ELD SHA update for the 23.x branch as well.

Left a comment on this--seems odd that the added `git switch` would be needed?

I guess it should be harmless, but curious why we need it.

jonathonpenix (approved):

jonathonpenix:
Also, before/when merging, please update the PR title/description to reflect whatever the final commit title/description will be.

What you currently have in the commit seems good to me

PR #480, opened 2026-07-07
[cpullvm][Github] Add workflow to reject commits with Gerrit Change-Id lines

jonathonpenix:
As folks transition from Gerrit to Github, we're seeing these Change-Id lines leak into changes they're pushing into Github.

They aren't needed, are just noise, and developers should get into the practice of not having these when they contribute to other Github-based projects (LLVM, Zephyr, etc.).

So, add a workflow that rejects any commits that have the Change-Id lines in their commit messages.

Assisted-by: claude


jonathonpenix:
An example of it working (rejecting commits with the Change-Id line) here: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/28903154791/job/85744385739?pr=481

lenary (approved):
LGTM

apazos (approved):

PR #494, opened 2026-07-12
[cpullvm] Add nightly Release build workflow

sriyalamar:
Add nightly-release.yml to build all 4 variants in Release mode (assertions disabled) by passing -DLLVM_ENABLE_ASSERTIONS=OFF via EXTRA_CMAKE_ARGS.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -21,6 +21,17 @@ name: Nightly Build and Test
 
 on:
   workflow_dispatch:
+    inputs:
+      enable_assertions:
+        description: 'Enable LLVM assertions'
+        type: boolean
+        default: true
+  workflow_call:
+    inputs:
+      enable_assertions:
+        description: 'Enable LLVM assertions'
+        type: boolean
+        default: true
   pull_request:
     paths:
       - '.github/workflows/nightly.yml'
@@ -28,6 +39,9 @@ on:
     - cron: '0 7 * * *' # 12 AM PT
 permissions:
   contents: read
+
+env:
+  EXTRA_CMAKE_ARGS: ${{ format('{0}', inputs.enable_assertions) == 'false' && '-DLLVM_ENABLE_ASSERTIONS=OFF' || '-DLLVM_ENABLE_ASSERTIONS=ON' }}
 
 jobs:
   build-and-test-Linux-x86:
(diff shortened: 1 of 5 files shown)


apazos (commented):
Is this ready for review?
Best to have the workflow be programmable than repeating the whole workflow.

.github/workflows/nightly.yml
```diff
@@ -48,6 +60,8 @@ jobs:
 
       - name: Build
         run: ./qualcomm-software/scripts/build.sh
+        env:
```
jonathonpenix: Is it possible to hoist these to the top level since we want the same env across all jobs/builds?

It'd save having to repeat it below I think

.github/workflows/nightly.yml
```diff
@@ -16,11 +16,23 @@
 #
 # Other build configurations such as building the runtimes on non-Linux
 # x86_64 platforms are built and tested separately.
+# Temporary: trigger nightly build to validate assertions OFF behavior.
```
jonathonpenix: You can remove this

The pull_request trigger as defined below will run the workflow if there's any changes to the file--so if you're already changing the file, comments like this aren't needed (they're only needed if there aren't any other changes but you want to trigger the workflow).

jonathonpenix (commented):
Left two small comments, but looks pretty close, thanks!

Would you also be willing to create a separate PR where we consistently handle the EXTRA_CMAKE_ARGS across all our build scripts? I think there's a few left after this that don't have the changes, but I think it makes sense to just standardize on that (I think we can use this for other things like sanitizers, etc.).

jonathonpenix:
Don't worry about the zizmor failure, that is something we need to sort out separately.

Please do check the alert that is generated on this PR though: https://github.com/qualcomm/cpullvm-toolchain/runs/89814027910?pr=494 (you can see it under the files changed tab, not sure if there's another way to view this)--I think we should probably address this if possible.

.github/workflows/nightly.yml
```diff
@@ -29,6 +40,9 @@ on:
 permissions:
   contents: read
 
+env:
+  EXTRA_CMAKE_ARGS: ${{ format('{0}', inputs.enable_assertions) == 'false' && '-DLLVM_ENABLE_ASSERTIONS=OFF' || '' }}
```
apazos: shouldn't the nightly have the assertions on, while nightly-release has it off?

sriyalamar: yes @apazos, The nightly build keeps assertions ON by default. The EXTRA_CMAKE_ARGS only activates when explicitly called from nightly-release.yml with enable_assertions: false, which turns assertions OFF for release builds only.

jonathonpenix (approved):

jonathonpenix:
(I'm just trying to test the cherry-pick thing to see if it is working, please ignore it--it is ok if we use the already open PR)

PR #497, opened 2026-07-13
[cpullvm][Github] Also reject PRs if Change-Id is found in the PR description

jonathonpenix:
check_change_id.yml used to only check the commits, so folks could update
their commit messages but not have that reflected in the PR description.
Reject if Change-Id is found in the PR description as well.

Assisted-by: claude

### DIFF
--- a/.github/workflows/check_change_id.yml
+++ b/.github/workflows/check_change_id.yml
@@ -1,22 +1,25 @@
 # Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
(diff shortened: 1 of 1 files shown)


jonathonpenix:
You can see this working (failing) here: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/29286534463/job/86940330306?pr=497 (I was just editing the description of this PR)

lenary (approved):
LGTM

svs-quic:
Did you mean "auto" reject in the commit title?

jonathonpenix:
I should have worded it better/more clearly, sorry--the thinking was something along the lines of it being a continuation of the PR/commit adding this workflow ("now _also_ reject if in the PR description in addition to commit messages").

But, I agree it isn't very clear from this PR alone though so I should have worded it differently.

PR #498, opened 2026-07-14
[riscv] Add Frame Pointers to Xqci configs

lenary:
The overheads here are:
- All libraries go from 893690 to 913660 bytes of rodata/text (+2.2%)
- libc.a specifically goes from 406804 to 420806 bytes of rodata/text (+3.5%)

This is a little larger than I expected, but we should have two things in our favour for the effect on user code:
1. Users don't link the entirety of our libraries (via linker GC)
2. Library code should be a small percentage of total application size

I think this trade-off is worth improved tracing.


lenary:
Across `embench-iot`, this adds on average 1% code size increase. The range is from around 0.8% to 1.5% increase across the suite.

quic-garvgupt (approved):

lenary:
I need to add this to the new xqci configs too.

lenary:
In talking to our customers:
- They are broadly only using the assembly routines in the libc, so this option won't change code size for them.
- They are using some C-implemented routines in libclang-rt, so this will help improve tracing for them there.

Frame pointers are a hard requirement for some of our customers so we are going to proceed with this.

quic-garvgupt (approved):

pgodeq:
LGTM as well.

pgodeq (approved):

lenary:
Backported to 23.x in #555

PR #500, opened 2026-07-14
[cpullvm] Update nightly build schedule to once daily at 12 AM PT

sriyalamar:
Changed the nightly workflow cron schedule from twice daily (12 AM and 12 PM PT) to once daily (12 AM PT only).


apazos (approved):
Remember to add it to the task list for release process.

PR #505, opened 2026-07-15
[cpullvm] remove test_exception_storage.pass.cpp from xfails.py

bojle:
This test failed due to ios_base::xalloc requiring an atomic function namely atomic_fetch_add_4. https://github.com/llvm/llvm-project/pull/208356 fixes it by removing dependence on atomic.


efriedma-quic (approved):
LGTM

jonathonpenix (approved):

PR #511, opened 2026-07-16
[DNM] Proof-of-concept picolibc-curr, picolibc-main overlays

jonathonpenix:
This adds two proof-of-concept picolibc overlays:
* `picolibc-main` tracking picolibc's main branch
    * Note that there's currently some XPASS-ing tests for this. Unclear to me if this is something new with mainline picolibc, or if there is an issue in one of these patches. Need to investigate further.
    * Note that the picolibc patch adding xqci memcpy/etc. is omitted. It needed enough rebasing that it didn't seem worth the time for the purposes of this PR.
* `picolibc-curr` that should be identical to our current `picolibc`

I don't think we actually want to merge this right now (we don't have a need for `picolibc-main`/`-curr` library). This is only intended to help test #540, #541, and #542 and show that those changes are sufficient for supporting this sort of thing.

---

Stacked on #542

### DIFF
--- a/.github/workflows/linux-premerge.yml
+++ b/.github/workflows/linux-premerge.yml
@@ -39,6 +39,14 @@ jobs:
             target_os: Linux-x86_64
             runner: cpullvm-ubuntu24-x86_64
 
+          # FIXME: this should probably go in the nightlies instead like
+          # musl-embedded, but I don't want to keep triggering those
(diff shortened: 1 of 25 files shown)


jonathonpenix:
Going to close this for now, will open a separate PR for the picolibc v1.8.12 overlay

PR #135, opened 2026-02-05
[NFC] Remove zero width space unicode

svs-quic:
They don't seem to show up in the commit diff but this is how git diff looks like locally for one of the files:

```
diff --git a/qualcomm-software/scripts/build_no_runtimes.sh b/qualcomm-software/scripts/build_no_run
times.sh
index 41ea4ac5928e..0bf9035ca793 100755
--- a/qualcomm-software/scripts/build_no_runtimes.sh
+++ b/qualcomm-software/scripts/build_no_runtimes.sh
@@ -5,7 +5,7 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
-# <U+200B><U+200B><U+200B><U+200B><U+200B>Changes from Qualcomm Technologies, Inc. are provided und
er the following license:
+# Changes from Qualcomm Technologies, Inc. are provided under the following license:
 # Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries. 
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

PR #512, opened 2026-07-16
[cpullvm][NFC] Remove additional zero width space unicode characters

jonathonpenix:
This should be the same idea as https://github.com/qualcomm/cpullvm-toolchain/pull/135

I see something like this if I open these files with vim:

<200b><200b><200b><200b><200b>Changes from Qualcomm Technologies

I didn't intentionally add these, so remove them.


jonathonpenix:
Failures should be unrelated

apazos (approved):

PR #518, opened 2026-07-20
[cpullvm] Switch musl-embedded to using the new versions.json SHA auto-update

jonathonpenix:
There are hardly any changes going into musl-embedded so it shouldn't make too much difference, but switch to using the SHA auto-update incase we do make any changes as it should again make bisecting a bit more consistent.


apazos (approved):

sriyalamar (approved):
LGTM

PR #519, opened 2026-07-20
[cpullvm] Install psutil to enhance test results reporting

pranav4330:
Installing psutil will impact our Windows builds (Arm64), and it will improve reporting for tests that time out so developers can investigate them more easily.

It is currently not working on windows x86 as there are many python versions being invoked which will be addressed in follow up patch.

python -m pip install ensures pip installs into the exact Python interpreter you're invoking, while pip install uses whichever python is first on your PATH, which may point to a different Python version unexpectedly.

Signed-off-by: Pranav Patil <pranpati@qti.qualcomm.com>

### DIFF
--- a/qualcomm-software/scripts/install_dependencies.ps1
+++ b/qualcomm-software/scripts/install_dependencies.ps1
@@ -9,3 +9,6 @@ pip install pyyaml
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
 pip install meson==1.10.0
+
+# Install psutil
+python -m pip install psutil


qualcomm-software/scripts/install_dependencies.ps1
navaneethshan, line 8: Add python -m here too.

jonathonpenix:
It might be worth rerunning the Windows builds and seeing what happens (our Windows builds don't timeout every run, but pretty often)

I still see this in the logs of the Windows x86 build, so I'm still not sure this is actually working like we hope/I'm guessing we'll see the same message if it does time out:

```
2026-07-21T01:16:08.5331972Z llvm-lit.py: D:\a\cpullvm-toolchain\cpullvm-toolchain\build\llvm\utils\lit\tests\lit.cfg:124: warning: Setting a timeout per test not supported. Requires the Python psutil module but it could not be found. Try installing it via pip or via your operating system's package manager.
```

Edit: yeah, there's a run already showing this actually--see this run from before the other non-psutil changes were pushed: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/29782127072/job/88485543954

The `python -m pip install psutil` lines are present, but we a) still time out and b) still see the message about no psutil

Although in https://github.com/qualcomm/cpullvm-toolchain/actions/runs/29782127072/job/88485543965 you can see the timeout in eld tests.

I wonder if there's two different pythons? 

From the x86 build that times out in the lit config:

note: using python: "c:/hostedtoolcache/windows/python/3.12.10/x64/python.exe"

But then in the final error about psutil:

Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.14.6\x64\Lib\threading.py", line 1082, in _bootstrap_inner

pranav4330:
@jonathonpenix 

as there are C:/hostedtoolcache/windows/Python/3.12.10/x64/python3.exe i think this is used by llvm-lit.

and 

 C:\hostedtoolcache\windows\Python\3.14.6\x64\python3.exe is used for installing psutil

from my understanding if install psutil into the same Python that lit is using 3.12.10 things would be more efficient.

it can be done using

C:\hostedtoolcache\windows\Python\3.12.10\x64\python.exe -m pip install psutil


Note:
2026-07-21T01:16:04.8261830Z llvm-lit.py: D:\a\cpullvm-toolchain\cpullvm-toolchain\build\_deps\eld-src\test\lit.cfg:600: note: using python: "c:/hostedtoolcache/windows/python/3.12.10/x64/python.exe"

navaneethshan:
@pranav4330, C:\hostedtoolcache\windows\Python\3.14.6\x64\python3.exe is also being used, did you do any experiment to find what python is being used and list all the python installations? I will spend some time on this.

qualcomm-software/scripts/install_dependencies.ps1
```diff
@@ -5,7 +5,12 @@
...
+python -m pip install pyyaml
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
-pip install meson==1.10.0
+python -m pip install meson==1.10.0
+
+
+#psutil instllation
+
+python -m pip install psutil
```
navaneethshan, line 14: Leave an empty line at the end.

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -12,3 +12,6 @@ sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-19
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
```
navaneethshan, line 13: drop this file, no need to update this file.

pranav4330:
@navaneethshan I  have made the changes

qualcomm-software/scripts/install_dependencies.ps1
```diff
@@ -5,7 +5,13 @@
...
 # Install pyyaml
-pip install pyyaml
+python -m pip install pyyaml
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
-pip install meson==1.10.0
+python -m pip install meson==1.10.0
+
+
+#psutil installation
```
navaneethshan: Always leave a space after # when you add a comment line.

\# Install psutil

pranav4330: my bad missed it Thank you for pointing it out. correction done

qualcomm-software/scripts/install_dependencies.ps1
```diff
@@ -5,7 +5,13 @@
...
+python -m pip install pyyaml
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
-pip install meson==1.10.0
+python -m pip install meson==1.10.0
+
+
+# psutil installation
+
+python -m pip install psutil
```
navaneethshan: don't leave empty line after comment, change comment to "Install psutil"

qualcomm-software/scripts/install_dependencies.ps1
```diff
@@ -5,7 +5,12 @@
...
 
 # Install pyyaml
-pip install pyyaml
+python -m pip install pyyaml
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
-pip install meson==1.10.0
+python -m pip install meson==1.10.0
+
+
```
jonathonpenix, line 12: Remove extra newline

qualcomm-software/scripts/install_dependencies.ps1
```diff
@@ -5,7 +5,12 @@
...
+# Install psutil
+python -m pip install psutil
+
```
jonathonpenix: Remove extra newline

jonathonpenix (commented):
Please update the PR/title description to reflect the changes in the PR/what the intent here is
- Remove reference to the temporary nightly changes that were removed

Other misc. thoughts:  
- Why the `python -m` changes--why is it needed? (or is it needed? it doesn't seem like it actually changed anything)?</br></br>Do other scripts need this same change/is someone going to handle this?
- Some mention of which targets this is expected to address would be good (this is only touching Windows, is there going to be a follow-up for Linux AArch64 and Linux x86 handled separately?)</br></br>Also, as we discussed earlier this isn't actually working for Windows x86 currently--some mention about that would be good

navaneethshan:
> Please update the PR/title description to reflect the changes in the PR/what the intent here is
> 
> * Remove reference to the temporary nightly changes that were removed
> 
> Other misc. thoughts:
> 
> * Why the `python -m` changes--why is it needed? (or is it needed? it doesn't seem like it actually changed anything)?Do other scripts need this same change/is someone going to handle this?
> * Some mention of which targets this is expected to address would be good (this is only touching Windows, is there going to be a follow-up for Linux AArch64 and Linux x86 handled separately?)Also, as we discussed earlier this isn't actually working for Windows x86 currently--some mention about that would be good

@apazos  was asking to get this merged before nightly build yesterday? if this is not making any difference to Aarch64 builds, then no point in merging.

pranav4330:
Hello @jonathonpenix ,

I have mentioned and addressed your comment,

as per why we need python -m pip install. it runs pip as a module of a specific python interpreter, so the package lands in the right environment.

if in future other scripts need this change we can address that issue as well.

jonathonpenix:
> > Please update the PR/title description to reflect the changes in the PR/what the intent here is
> > 
> > * Remove reference to the temporary nightly changes that were removed
> > 
> > Other misc. thoughts:
> > 
> > * Why the `python -m` changes--why is it needed? (or is it needed? it doesn't seem like it actually changed anything)?Do other scripts need this same change/is someone going to handle this?
> > * Some mention of which targets this is expected to address would be good (this is only touching Windows, is there going to be a follow-up for Linux AArch64 and Linux x86 handled separately?)Also, as we discussed earlier this isn't actually working for Windows x86 currently--some mention about that would be good
> 
> Ana Pazos (@apazos) was asking to get this merged before nightly build yesterday? if this is not making any difference to Aarch64 builds, then no point in merging.

@navaneethshan Sorry, I might have worded this poorly. https://github.com/qualcomm/cpullvm-toolchain/pull/519#issuecomment-5030353463 is still basically the limit of my understanding at this point. My understanding of our builds:
- Windows AArch64 -> we see some change with this PR
- Windows x86 -> in theory should be impacted by this change (install_dependencies.ps1 is used by this build too), but this change isn't actually working on this builder for some reason
- Linux AArch64 / Linux x86 -> not touched by this change

So I think/agree there's value in merging _a_ change to handle Windows AArch64 right now since we did see it a) have an effect and b) timeout an eld test--that way someone can look at said test.

But given that it doesn't really seem like we have a good handle on what is going on (see: Win x86 build comment above), I'd prefer us keep this change as narrow as possible to achieve that goal.

> Hello Jonathon Penix (@jonathonpenix) ,
> 
> I have mentioned and addressed your comment,
> 
> as per why we need python -m pip install. it runs pip as a module of a specific python interpreter, so the package lands in the right environment.
> 
> if in future other scripts need this change we can address that issue as well.

@pranav4330 https://github.com/qualcomm/cpullvm-toolchain/pull/519#issuecomment-5030353463 is still basically the limit of my understanding at this point--if I'm getting something wrong here, please correct me. Given that, the PR title/description still needs to be updated:
- This PR _doesn't_ work for Windows x86 currently
- native_nightly file isn't changed
- install_dependency.sh isn't changed

Re: `python -m pip`, please include in the description (which will be used as the commit message when merging; these don't need to be long explanations but at least *some* mention):
* Why `python -m` form is preferred
* That this change modifies (and why) the non-psutil dependencies in the same way
* What this implies for our other scripts that still use just `pip`

Given that IMO we don't have a good handle on what is actually working here and why (see: Win x86 build) I'd prefer we keep this change as narrow as possible or at least be explicit about the changes we are making and why.

I realize the non-psutil changes were explicitly requested above so I'll defer to you and @navaneethshan to decide if we keep these changes , but we should at least be clear in the PR description/title (eventual commit message/title) about the changes we're making and why.

jonathonpenix:
Updating your PR branch should resolve the build failure

apazos:
@pranav4330 
- the only line change we should see is 
python -m pip install psutil
no other change please
- you need to fix PR title and description/summary (first text box) to explain the change. Note it impacts both Win on x86 and but Win on Arm64, by improving the reporting of tests that time out, and state that on Win on x86 there are more issues due to different python versions being invoked which will be addressed in follow up patch.

apazos (commented):
Please
 - fix long PR title, maybe just "Install psutil to enhance test results reporting"
 - Need proper punctuation in the PR description (out".", missing end point after (patch"), and the last paragraph you use pip instead of python. Re-read what you wrote and fix them please.

apazos (approved):

jonathonpenix (approved):
Needs rebase to get clean builders before merging.

pranav4330:
yes rebase is done can you merge. @jonathonpenix

PR #523, opened 2026-07-21
[cpullvm][Github] Update QC preflight checks to v2.0.8 release

jonathonpenix:
This release contains fixes for a few things related to our repo (Apache license auto-detection, skipping email checks on pushes). So, update to this version to get these changes.

### DIFF
--- a/.github/workflows/qcom-preflight-checks.yml
+++ b/.github/workflows/qcom-preflight-checks.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   preflight:
     name: Run QC Preflight Checks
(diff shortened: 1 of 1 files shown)


apazos (approved):

jonathonpenix:
As a late realization while looking at unrelated stuff, I guess v2 is a mutable tag so we could have stayed. But, pointing at the specific SHA/version aligns with Github/LLVM guidance and what we do elsewhere in the repo.

So I think the change is still correct, just the wording is misleading (we aren't really upgrading, more just rewriting the tag)

Sorry about the mixup on my end

PR #533, opened 2026-07-22
[multilib] Add Xqccmt to Xqci Variant

lenary:
This was missing in 22.x because we didn't have support for Xqccmt in the toolchain. The intention was that it was there, so add it now.


lenary:
This is intended to be back-ported to 23.x

jonathonpenix (approved):
LGTM in terms mechanics of adding the extension, will leave it to others to confirm said extension should be added 😃

Just to double check (I don't have an opinion/preference either way): I think the xqci linux variant we currently have also doesn't have xqccmt, are we leaving it that way?

lenary:
> Just to double check (I don't have an opinion/preference either way): I think the xqci linux variant we currently have also doesn't have xqccmt, are we leaving it that way?

D'oh. No, I need to update the linux variant too.

lenary:
@navaneethshan I think we're waiting for the eld fix on mainline, I guess it should come in from automerge shortly? Not sure. This commit: https://github.com/qualcomm/eld/commit/6c1c37f32017d852d54d44c24c0e7e6566c35082

pgodeq:
IMO, LGTM in terms of extensions, we had reviewed.

jonathonpenix:
> Navaneeth Shanmugasundaram (@navaneethshan) I think we're waiting for the eld fix on mainline, I guess it should come in from automerge shortly? Not sure. This commit: [qualcomm/eld@6c1c37f](https://github.com/qualcomm/eld/commit/6c1c37f32017d852d54d44c24c0e7e6566c35082)

FWIW, the non-cpullvm/llvm repos are updated on the same schedule as automerge runs, see [here](https://github.com/qualcomm/cpullvm-toolchain/blob/b289d3e4bed08d1ea8c21835d43e6d2d483e22e6/.github/workflows/automerge.yml#L45-L54) (so every 30 min)

At this point if you update your branch it should have the change

jonathonpenix (approved):

PR #534, opened 2026-07-23
[cpullvm] Don't include cmake-bridge.cfg for libc++ tests

jonathonpenix:
This seems to be unneeded after [1] and causes failures if left in.

Follow the recommendation in the PR about no longer including `cmake-bridge`.

[1] https://github.com/llvm/llvm-project/commit/df4bb1f3106d39e5c6b767a28c0531efbbebc2d2


jonathonpenix:
Needed by/tested in https://github.com/qualcomm/cpullvm-toolchain/pull/532

apazos (approved):

PR #538, opened 2026-07-23
[multilib] Add RISC-V control-flow protection flags

lenary:
Since https://github.com/llvm/llvm-project/pull/205202, clang allows us to choose different libraries based on the amount of control flow protection requested.

None of our libraries are built with control flow protection, so we make this explicit, and add the `-fcf-protection=none` to the multilib matching flags.

This ensures that users get a warning if they try to request control flow protection with our libraries.


jonathonpenix (approved):

lenary:
(I think this does want backporting to 23.x, and applying to the new configs coming in concurrently, which I can update after they land)

apazos (approved):
Thanks for pushing this change warning users that libs were not built with CFI enabled.

jonathonpenix:
Sorry for late comment, but maybe it'd be worth picking on one variant and adding a test in our multilib tests that a non-none produces the warning/not matching?

qualcomm-software/test/multilib/riscv32.test
```diff
@@ -79,3 +79,15 @@
...
 # NOT-FOUND: warning: no multilib found matching flags
+
+
+## Check that we only find libraries for `-fcf-protection=none`
+
+# RUN: %clang -print-multi-directory --target=riscv32-unknown-elf -march=rv32imac -mabi=ilp32 -fno-pic -fcf-protection=none | FileCheck %s --check-prefix=FCF-FOUND
+# FCF-FOUND: riscv32-unknown-elf/riscv32imac_ilp32_nopic{{$}}
+# FCF-FOUND-EMPTY:
+
+# RUN: %clang -print-multi-directory --target=riscv32-unknown-elf -march=rv32imac -mabi=ilp32 -fno-pic -fcf-protection=branch | FileCheck %s --check-prefix=FCF-NOT-FOUND
```
jonathonpenix: Think these are just missing the 2>&1?

lenary: Thanks, fixed.

jonathonpenix (approved):

jonathonpenix (approved):

PR #539, opened 2026-07-23
[linux] Clang Config Files for Linux Variants

lenary:
This should make it easier to run the linux variants without having to use shell script wrappers.

I chose to do one config file per variant, which can cover C and C++, alternatively I could do separate ones that leave the C++ parts out of the configurations for C.

---

Written with the assistance of AI.


qualcomm-software/scripts/build_linux_runtimes.sh
```diff
@@ -488,4 +497,49 @@ for VARIANT in "${VARIANTS[@]}"; do
...
+    echo "\$-Wl,--no-as-needed"
+    echo "\$<CFGDIR>/${VARIANT_STDLIB_REL}/lib/crt1.o"
+    echo "\$<CFGDIR>/${VARIANT_STDLIB_REL}/lib/crti.o"
+    echo "\$<CFGDIR>/${VARIANT_STDLIB_REL}/lib/crtn.o"
+    echo "\$-Wl,--start-group"
+    echo "\$-lc"
+    echo "\$-lclang_rt.builtins"
+    echo "\$-Wl,--end-group"
+    echo "\$-Wl,--defsym,__dso_handle=0"
+  } > "${BASE_INSTALL_DIR}/cfgs/${VARIANT}.cfg"
```
jonathonpenix: My understanding is we'll end up with something like `rv32imaf_zve32f_zvfh_zba_zbb_ilp32f.cfg`. Is it worth appending something to make it clear that this is a Linux config?

More magic naming conventions aren't great, but it might help a) disambiguate between ex: the baremetal `riscv32imaf_zve32f_zvfh_zba_zbb_ilp32f_nothreads` (or, whatever the embedded variants corresponding to the particular Linux config are) and b) might make matching the configs when in copy_target_libraries.py marginally easier?

qualcomm-software/cmake/copy_target_libraries.py
```diff
@@ -102,6 +119,15 @@ def main():
             for folder in linux_lib_folders:
                 move_folder(os.path.join(tmp, "*", folder), linux_lib_dir)
 
+            # Move the per-variant clang configuration files. These live in
+            # the distribution's bin/ directory alongside musl-embedded.cfg
+            # (which is not a linux-runtime variant cfg and must stay put).
+            move_files(
+                os.path.join(tmp, "*", "bin", "*.cfg"),
+                os.path.join(linux_lib_dir, "cfgs"),
+                exclude_names={"musl-embedded.cfg"},
```
jonathonpenix: Not a huge deal since I'm not sure how many embedded libcs we'll end up supporting, but maybe it'd be worth seeing if there's a common thing we can add into the linux cfgs to distinguish them from the embedded ones?

More "magic" naming conventions aren't great, but I guess it wouldn't suprise me if someone (probably me) forgot to update this when adding ex: a new picolibc variant, and we start copying that around too. (which should be harmless, but a bit odd).

Left another similar comment about this below

jonathonpenix (commented):
Minor question about naming, but otherwise LGTM I think

qualcomm-software/scripts/build_linux_runtimes.sh
lenary: Maybe we add `musl_` to the start, to give us: `musl_rv32imaf_zve32f_zvfh_zba_zbb_ilp32f.cfg`

I'll do this as I think it fixes the issue below too.

jonathonpenix (approved):

PR #542, opened 2026-07-24
[cpullvm] Support additional picolibc versions as overlays

jonathonpenix:
This should allow us to support multiple different versions of picolibc at
the same time. These are expected to be added roughly like a new libc named
with the 'picolibc' prefix. There are a few exceptions to these new picolibc
versions being handled as a "full" new libc to save duplication:
* Library variants supported by 'picolibc' in json are considered supported by
  all picolibc versions.
* All the build flags (picolibc's base build flags, the flags used by libc++
  when building against picolibc, json library flags etc.) are shared between
  versions.
* xfails.py applies the same xfails for all picolibc versions.

All of these can be changed/overridden as the need arises, but this seemed like
a safe-ish place to start.

Notably though, the base 'picolibc' is still treated as a 'default' library, so
it is disallowed from being in an overlay, has the version file named slightly
differently, etc. This special handling isn't extended to the picolibc version
variants.

Assisted-by: claude

### DIFF
--- a/qualcomm-software/test/multilib/aarch64.test
+++ b/qualcomm-software/test/multilib/aarch64.test
@@ -3,44 +3,44 @@
 # Should also revisit whether there are relevant tests we can reuse even without
 # the original multilib.
 
-# RUN: %clang_with_sysroot -print-multi-directory --target=aarch64-none-elf | FileCheck %s --check-prefix=CHECK-AARCH64
-# RUN: %clang_with_sysroot -print-multi-directory --target=aarch64-none-elf -munaligned-access | FileCheck %s --check-prefix=CHECK-AARCH64
-# RUN: %clang_with_sysroot -print-multi-directory --target=aarch64-none-elf -mcpu=cortex-a53 | FileCheck %s --check-prefix=CHECK-AARCH64
+# RUN: %clang_with_libc_cfg -print-multi-directory --target=aarch64-none-elf | FileCheck %s --check-prefix=CHECK-AARCH64
+# RUN: %clang_with_libc_cfg -print-multi-directory --target=aarch64-none-elf -munaligned-access | FileCheck %s --check-prefix=CHECK-AARCH64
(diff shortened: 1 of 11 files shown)


qualcomm-software/test/multilib/aarch64.test
```diff
@@ -3,44 +3,44 @@
 # Should also revisit whether there are relevant tests we can reuse even without
 # the original multilib.
 
-# RUN: %clang -print-multi-directory --target=aarch64-none-elf | FileCheck %s --check-prefix=CHECK-AARCH64
-# RUN: %clang -print-multi-directory --target=aarch64-none-elf -munaligned-access | FileCheck %s --check-prefix=CHECK-AARCH64
-# RUN: %clang -print-multi-directory --target=aarch64-none-elf -mcpu=cortex-a53 | FileCheck %s --check-prefix=CHECK-AARCH64
+# RUN: %clang_with_sysroot -print-multi-directory --target=aarch64-none-elf | FileCheck %s --check-prefix=CHECK-AARCH64
```
jonathonpenix: Think at one point I discussed with Sam about turning this into essentially a %clang_with_cfg or just overriding %clang to also include the cfg.

Will tinker with this

jonathonpenix:
Rebasing to unstack #540, #541 as they've been merged.

qualcomm-software/test/multilib/aarch64.test
jonathonpenix: Addressed this in https://github.com/qualcomm/cpullvm-toolchain/pull/542/changes/2bb45a05b2803b2b17c69d1ded6a699b4cbc35cc

It requires changing where we place the config file in builds, but I think that should be ok. It's also a bit different in that `%clang_with_sysroot` == `%clang` for our main libc (since we don't generate a config file for that, but I think that is ok--we're closer to mirroring/testing what our users will be doing.

apazos (approved):

PR #550, opened 2026-07-26
[cpullvm] Add nightly Release build workflow 23.x

sriyalamar:
Add nightly-release.yml to build all 4 variants in Release mode (assertions disabled) by passing -DLLVM_ENABLE_ASSERTIONS=OFF via EXTRA_CMAKE_ARGS.


apazos (commented):
Can we explicitly set Assertions to ON and OFF and avoid relying on default behavior?

PR #552, opened 2026-07-27
23.x: [multilib] Add Xqccmt to Xqci Variant

lenary:
This was missing in 22.x because we didn't have support for Xqccmt in the toolchain. The intention was that it was there, so add it now.

(cherry picked from commit 4329ae4259f685281ad12401683484c084f5cf2d)


jonathonpenix (approved):

lenary:
I'm not entirely sure the impact of this on internal testing.
- We need the `musl` patch for Pankaj's work.
- Our internal wrappers will still choose the same linux multilib as before
- The qemu being used by those test configurations doesn't support xqccmt yet, and we haven't added `xqccmt=true` to the cpu arguments (yet).
- We haven't enabled `--relax-tbljal=xqccmt` in that config (yet), which should prevent relaxations involving those instructions.

So I think this is ok to merge? I'm not 100% sure.

pgodeq (approved):
From cpullvm side, I think with this, we would have all the ground-work to enable test config for new Xqci Linux variants. 
As picolibc tests pass, so I think, we should be good. 
We need the new qemu wrappers and .ini files, to further validate the linux variant.

lenary:
The two internal testing runs I did have worked, presumably because they aren't passing `--relax-tbljal=xqccmt`, so don't yet have xqccmt instructions.

I'm going to merge this to unblock the other backport PR, and then focus on HF config changes.

PR #555, opened 2026-07-29
23.x: [riscv] Add Frame Pointers to Xqci configs

lenary:
The overheads here are:
- All libraries go from 893690 to 913660 bytes of rodata/text (+2.2%)
- libc.a specifically goes from 406804 to 420806 bytes of rodata/text (+3.5%)

This is a little larger than I expected, but we should have two things in our favour for the effect on user code:
1. Users don't link the entirety of our libraries (via linker GC)
2. Library code should be a small percentage of total application size

(cherry picked from commit a756c78f8bb741d241d0af5c02137b054ac1b40b)


pgodeq (approved):
LGTM

lenary:
Internal testing runs of both riscv32 and riscv64 came back clean.

jonathonpenix (approved):

PR #556, opened 2026-07-29
[zephyr] zephyr twister test report for easy comparison

pranav4330:
Updated zephyr-twister-tests.yml to compare the previous and latest test runs and generate a consolidated report for easier analysis.
Added generate_twister_report.py to generate a detailed Excel (.xlsx) report.

Signed-off-by: Pranav Patil pranpati@qti.qualcomm.com

### DIFF
--- a/.github/actions/generate_twister_report.py
+++ b/.github/actions/generate_twister_report.py
@@ -1,0 +1,332 @@
+#!/usr/bin/env python3
+"""
+generate_twister_report.py
+──────────────────────────
+Compare two Zephyr Twister JUnit XML files (previous and current CI run) and
+produce a formatted .xlsx report.
+
+Usage:
+    python generate_twister_report.py \
+        --prev  previous_twister.xml \
+        --curr  merged_twister.xml   \
+        --out   twister_comparison_report.xlsx \
+        [--prev-date "Jun 14"] \
+        [--curr-date "Jun 15"]
+
+The script is also called automatically by the 'Zephyr Twister Tests' GitHub
+Actions workflow; the workflow passes the correct paths and date labels.
+"""
+
+import argparse
+import sys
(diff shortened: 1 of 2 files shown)


apazos:
Please follow English grammar when writing PR description which will become the commit message of the merged PR. You can use co-pilot for that.

Suggestions:
Updated zephyr-twister-tests.yml to compare the previous and latest test runs and generate a consolidated report for easier analysis.
Added generate_twister_report.py to generate a detailed Excel (.xlsx) report.

If the code generated with the help of AI tooling, you need to add a statement following LLVM code standards: Assisted-by: commit trailer or a brief note in the PR description. And make sure you understand all that the code is doing.

jonathonpenix:
A few high level comments (in addition to Ana's), I'll try to take a closer look at this later in the week:
- Zephyr tests should be running again (successfully finding a toolchain)--please update your branch so we can see this running/look at what the generated sheet looks like.
- I don't know if others have strong preferences, but, just skimming this, the vast majority of this PR seems to be dealing with creating the spreedsheet itself rather than doing any processing of the xml files. Do we actually need this as a spreadsheet/does this format buy us anything for having to maintain it?</br></br>If we can dramatically simplify things by reducing what we're outputting (even just dumping to a text file) it seems like that might be worthwhile</br></br>Two thoughts along these lines:
    - Information about what is passing doesn't seem useful to me, my expectation is that people would mostly want to look at what is failing, which is/should be a relatively small number of things.
    - Right now, Zephyr tests are a bit messy but in theory they should be getting cleaner over time. As that happens, it seems like very simple summary information might be more helpful than a more elaborate setup. See ex: Zephyr's [test summaries](https://github.com/zephyrproject-rtos/zephyr/actions/runs/31421733095).

PR #560, opened 2026-07-30
[cpullvm][musl] Switch musl repository from https to git protocol

pranav4330:
The HTTPS endpoint for the musl repository is frequently inaccessible, which can prevent fetching sources. 

The git protocol endpoint (git://git.musl-libc.org/musl) remains available and cloning works
reliably, so switch to using it instead.

### DIFF
--- a/qualcomm-software/versions.json
+++ b/qualcomm-software/versions.json
@@ -11,7 +11,7 @@
       "tag": "01254932e8e81085817ed61fd858648584ffe37c"
     },
     "musl": {
-      "url": "https://git.musl-libc.org/git/musl",
+      "url": "git://git.musl-libc.org/musl",
       "tagType": "tag",
       "tag": "v1.2.5"
     },


qualcomm-software/versions.json
```diff
@@ -11,7 +11,7 @@
       "tag": "01254932e8e81085817ed61fd858648584ffe37c"
     },
     "musl": {
-      "url": "https://git.musl-libc.org/git/musl",
+      "url": "git://git.musl-libc.org/git/musl",
```
jonathonpenix: This doesn't look like the correct URL, see https://git.musl-libc.org/cgit/musl

pranav4330: The https:// web interface is down (Bad Gateway), while the git:// protocol talks directly to the Git daemon on the server, bypassing the web server entirely.

jonathonpenix: The point was that the link that was used for the git interface wasn't right.

pranav4330: Yes, I made the change in the link, and the build is progressing successfully now. Previously, it was failing.

navaneethshan:
Fix the commit message and typo. The issue is not that the repository is "currently" unavailable - the HTTPS git URL is frequently inaccessible, which is why we are switching..

pranav4330:
@navaneethshan addressed.

navaneethshan (approved):
LGTM

PR #567, opened 2026-08-03
[cpullvm][Github] Use latest successful Linux x86 build in Zephyr tests

jonathonpenix:
Previously we'd look for the latest successful nightly.yml build, but our
Windows builders are fairly unreliable. So, ex: Zephyr tests aren't able
to find a toolchain to run with, even if the Linux x86 builds in these
(overall failing) nightlies were successful.

To try and alleviate this, by default check only if the particular toolchain
we're looking for completed successfully.

Assisted-by: claude


jonathonpenix:
This seems to be working correctly--the runs are still going, but they're pulling a toolchain as expected: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/30841998946/job/91781041556?pr=567

(they're pulling run id=30796332937, number=487 which is https://github.com/qualcomm/cpullvm-toolchain/actions/runs/30796332937 where the Linux build succeeded but the overall run timed out)

apazos (approved):

pranav4330 (approved):

PR #568, opened 2026-08-03
[cpullvm] Remove outdated rv64 xfails

jonathonpenix:
This is now xpass-ing: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/30721037476/job/91424782151#step:6:31575

Unclear why, but it is also unclear why we needed to xfail it in the first
place.

### DIFF
--- a/.github/workflows/libcxx-test.yml
+++ b/.github/workflows/libcxx-test.yml
@@ -9,7 +9,6 @@
 
 name: Run libc++ tests
 
-# DNM: temp change to trigger workflow
 on:
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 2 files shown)


apazos (approved):

jonathonpenix:
Two runs with this have passed: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/30842640817?pr=568 and https://github.com/qualcomm/cpullvm-toolchain/actions/runs/30842640817/attempts/1?pr=568

If this proves to be intermittent pass/fail, we shouldn't be xfailing it anyway but marking it unsupported (and investigating it at some point).

So for now, I think it it seems ok to go forward with this.

jonathonpenix:
This is also xpassing in ex: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/30960911381/job/92164466464 (picolibc overlay patches)

apazos (approved):

PR #578, opened 2026-08-05
[cpullvm][Github] Remove eld support patch application from Zephyr tests

jonathonpenix:
The patch is now upstream so we don't need to add this as a patch anymore: https://github.com/zephyrproject-rtos/zephyr/commit/8925ae2806dcae57275872fbe95f93d948e1d039


apazos (approved):

jonathonpenix:
Going to go ahead and merge overtop of the failures. The Linux premerge failure is an LLVM failure--this patch is totally unrelated/can't cause that sort of thing and our builders are overwhelmed as it is.

The Zephyr tests are working as expected (building against eld)

PR #579, opened 2026-08-05
[cpullvm] Don't clone libc repos when using prebuilt libraries

jonathonpenix:
Currently, when `PREBUILT_TARGET_LIBRARIES` is used, the various libc repos (some combination of picolibc, musl, and musl-embedded depending on what is enabled) are still cloned. They aren't used to build the runtimes, but are needed to get version information and license files. This is a bit unfortunate as:
* We've had instability with some repos (musl) and see builds that fail when cloning these repos, even though the repo is only used for version/license info.
* It seems more correct to have the final version information/licenses that are installed directly match that of the copied runtimes (though in practice this isn't really an issue since we use SHAs/tags that don't tend to change often).

So, if `PREBUILT_TARGET_LIBRARIES` is used, don't clone the relevant libc repos and instead derive the license/version information directly from the package we're copying runtimes from.

Note that this support is written assuming the package being copied from should have been configured in basically the same way as what is being built. So if someone tries to ex: build with `musl-embedded` configured as the main library but copy out of a package built with `picolibc` as  the main library, things are going to break (even though all the required information is technically present if ex: Linux libraries are turned on).

Assisted-by: claude


jonathonpenix:
Note to self: I removed this from the stack in https://github.com/qualcomm/cpullvm-toolchain/pull/581 to make it easier to merge things.

Once that goes in, there needs to be additional fixups to handle the licenses situations for the different picolibc versions (check https://github.com/jonathonpenix/cpullvm-toolchain/tree/bak/license_copy_picolibc_1812_overlay)

jonathonpenix:
Successful nightly run (which copies the runtimes) with these changes: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/32053537901/job/95458378347?pr=579

I downloaded the artifacts and the VERSION.txt/THIRD-PARTY-LICENSES.txt/third-party-licenses/ files/folders look right.

Locally, I looked at builds with picolibc, picolibc-v1812, and musl-embedded as the main C library with and without Linux libraries enabled and things looked ok.

navaneethshan (approved):
LGTM

Yethendra (approved):
LGTM

PR #582, opened 2026-08-07
[ATfE] Remove PICOLIBC_TEST hack (#666)

jonathonpenix:
The minimum version of meson required by picolibc has been increased to 0.61, and so the PICOLIBC_TEST hack to enable builds on older versions has been removed from upstream test scripts. The scripts used by ATfE still expected this variable to be set, and so this has now also been removed to match.

In order to continue to set stdin to /dev/null, this is now done in the run_qemu script.

(cherry picked from commit 779e73e74ff8c9c15fe5dd95e046c9665655d580)

---

This is a cherry-pick from ATfE [1]. Our "main" picolibc doesn't need this, but this will be needed by our soon-to-be-added picolibc v1.8.12 overlay as [2] is present in that version (which removes the support for `PICOLIBC_TEST`).

This is safe for us to do in cpullvm with picolibc versions before the meson minimum requirement bump in picolibc [3] as cpullvm needs meson >= 1.9.0 anyway (for eld support in meson).

[1] https://github.com/arm/arm-toolchain/commit/779e73e74ff8c9c15fe5dd95e046c9665655d580
[2] https://github.com/picolibc/picolibc/commit/45dde9042ad91c54e360d2866904e7ec9f74128b
[3] https://github.com/picolibc/picolibc/commit/33065fe2705b5e9211177df09df21ee7aebe2dc2


lenary (approved):
Yes please remove this awful hack.

PR #584, opened 2026-08-07
23.x: [cpullvm] Add picolibc v1.8.12 overlay

jonathonpenix:
Add a new C library variant for picolibc v1.8.12 (`picolibc-v1812`) and start building it as an overlay.

For now, this shares all the same variants as the main `picolibc` library.

picolibc v1.8.12 includes API-breaking changes to `off_t`, so this is intended to serve users on codebases that expect such changes (Zephyr, for example seems to getting ready to make changes to support picolibc v1.8.12).

---

Stacked on #582 #583

I'll bring #582 and #583 to the branch once they're merged in `qualcomm-software`.

This also needs #595--I'll again bring that to the branch once they're merged in `qualcomm-software`.

This also also needs #598 for everything to pass.

I'm not going to stack #595 and #598 at this point, will wait for things to start merging.

### DIFF
--- a/.github/workflows/linux-premerge.yml
+++ b/.github/workflows/linux-premerge.yml
@@ -39,6 +39,14 @@ jobs:
             target_os: Linux-x86_64
             runner: cpullvm-ubuntu24-x86_64
 
+          # FIXME: this should maybe go into the nightly instead,
+          # but I don't want to keep triggering those until we're
+          # closer to being ready to merge.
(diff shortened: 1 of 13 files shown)


jonathonpenix:
Going to close this--I removed the extra stacking in the mainline commit, so I'll just cherry-pick to the branch directly rather than rebase, etc.

PR #586, opened 2026-08-07
[xfails][NFC] Clearer description of the "Insufficient RAM" xfail

bojle:


jonathonpenix:
I went ahead and merged https://github.com/qualcomm/cpullvm-toolchain/pull/595 which removes this xfail. So I think we can close this.

Even so, thanks a lot @bojle for opening/working on this--seeing the extra info was very helpful in triaging some semi-related picolibc failures we saw elsewhere.

PR #588, opened 2026-08-10
Disable release/23.x auto-sync to preserve RC3 baseline

navaneethshan:
Stop automated upstream synchronization for release/23.x to avoid pulling changes beyond the RC3 release candidate. Remaining RC3 updates will be synced manually as needed.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -19,8 +19,6 @@ jobs:
         branches:
           - from_branch: main
             to_branch: qualcomm-software
-          - from_branch: release/23.x
-            to_branch: release/qualcomm-software/23.x
     steps:
       - name: Configure Access Token
         uses: actions/create-github-app-token@f45685208fd9b88321d74015b5996fc8c3e43d18 # v1.0.0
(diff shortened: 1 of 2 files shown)


jonathonpenix:
In terms of motivation for the patch, I guess what's the advantage of doing this? My understanding is we want to follow community up to their 23.1.0 release (not stopping early like we've done in the past for some internal releases). So re: "Remaining RC3 updates will be synced manually as needed." my understanding is we'll want all these commits.

Depending on the above (if the goal is to freeze the toolchain for testing or something temporarily) do we need to be stopping auto merging for ex: eld as well?

navaneethshan:
RC3 is expected tonight (8/11), IIRC, we discussed that we want to pick ONLY upto rc3 for the release?
"Remaining RC3 updates will be synced manually as needed." - if we stop now, and if few more patches gets added before rc3 tonight, then we will sync only those few patches onto our branch. I am not sure if we have to peg eld as well, @apazos ?

apazos:
The final llvm.org update to the release branch is on August 25th, I think we should pick up all patches till then, until they tag 23.1.0. Might be worth stopping a day before this date to ensure we do not go beyond 23.1.0 commits.

navaneethshan:
OK, closing this PR. @sriyalamar is working on WF update to sync the branch upto an upstream tag, so we don't have to disable manaully once that change merges.

PR #593, opened 2026-08-12
[cpullvm] Add stop_at_tag for sync_from_upstream workflow

sriyalamar:
Add stop_at_tag logic to cap any release branch.
Syncs normally until the tag appears upstream, then freezes the branch at that commit. Subsequent runs are no-ops. Also added tag for release/23.x to stop at llvmorg-23.1.0.

### DIFF
--- a/.github/workflows/sync_from_upstream.yml
+++ b/.github/workflows/sync_from_upstream.yml
@@ -12,7 +12,10 @@ jobs:
     if: github.repository == 'qualcomm/cpullvm-toolchain'
     strategy:
       matrix:
-        branch: [main, release/23.x]
+        include:
+          - branch: main
+          - branch: release/23.x
+            stop_at_tag: llvmorg-23.1.0
     env:
       UPSTREAM_REMOTE: https://github.com/llvm/llvm-project.git
     steps:
@@ -27,13 +30,34 @@ jobs:
         with:
           ref: ${{ matrix.branch }}
           token: ${{ steps.generate-token.outputs.token }}
+          fetch-depth: 0
       - name: Configure Upstream Remote
         run: git remote add upstream ${{ env.UPSTREAM_REMOTE }}
       - name: Fetch Upstream Changes
-        run: git pull --ff-only upstream ${{ matrix.branch }}
+        run: |
+          git pull --ff-only upstream ${{ matrix.branch }}
+          if [ -n "${{ matrix.stop_at_tag }}" ]; then
+            git fetch upstream "refs/tags/${{ matrix.stop_at_tag }}:refs/tags/${{ matrix.stop_at_tag }}" 2>/dev/null || true
+          fi
       - name: Configure Git Identity
         run: |
           git config --local user.name "github-actions[bot]"
           git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
       - name: Push Changes
-        run: git push origin ${{ matrix.branch }}
+        run: |
+          STOP_TAG="${{ matrix.stop_at_tag }}"
+          if [ -n "$STOP_TAG" ] && \
+             git rev-parse "$STOP_TAG" > /dev/null 2>&1 && \
+             git merge-base --is-ancestor "$(git rev-parse "${STOP_TAG}^{}")" HEAD; then
+            STOP_COMMIT=$(git rev-parse "${STOP_TAG}^{}")
+
+            if [ "$(git rev-parse HEAD)" = "$STOP_COMMIT" ]; then
+              echo "Already at $STOP_TAG ($STOP_COMMIT), nothing to sync."
+              exit 0
+            fi
+
+            echo "Freezing ${{ matrix.branch }} at $STOP_TAG ($STOP_COMMIT)."
+            git push --force-with-lease origin "${STOP_COMMIT}:refs/heads/${{ matrix.branch }}"
+          else
+            git push origin ${{ matrix.branch }}
+          fi


jonathonpenix (commented):
In general I think this looks good, just a few misc. questions about how this works/why we're doing it this way:

- My understanding is this is to help with stopping our automerge at a certain point. Is there an advantage to controlling this from the upstream sync side rather than the automerge side?</br></br>Controlling it from the automerge side seems like it'd have the benefit of keeping our upstream mirror branches as upstream mirrors. And then that way controlling automerge is done through it's own code, not relying on the upstream sync doing a particular thing.</br></br>Maybe not a big deal, mostly curious
- Does this work if tags are created after we move past the tagged point?</br></br>I'm assuming LLVM will generally not have that happen, but not entirely sure if there's a guarantee. Or, if ex: a tag is changed. Do we at least get an error or would it push weird things?
- If someone ex: mistakenly specifies a tag on a different branch than the one listed, does the push fail or do we get a weird result?

sriyalamar:
> In general I think this looks good, just a few misc. questions about how this works/why we're doing it this way:
> 
> * My understanding is this is to help with stopping our automerge at a certain point. Is there an advantage to controlling this from the upstream sync side rather than the automerge side?Controlling it from the automerge side seems like it'd have the benefit of keeping our upstream mirror branches as upstream mirrors. And then that way controlling automerge is done through it's own code, not relying on the upstream sync doing a particular thing.Maybe not a big deal, mostly curious
> * Does this work if tags are created after we move past the tagged point?I'm assuming LLVM will generally not have that happen, but not entirely sure if there's a guarantee. Or, if ex: a tag is changed. Do we at least get an error or would it push weird things?
> * If someone ex: mistakenly specifies a tag on a different branch than the one listed, does the push fail or do we get a weird result?

@jonathonpenix, Chosen sync-side because, I thought release/23.x exists solely to feed automerge and we do not use it anywhere else and once we reached tag, we can remove 23.x in both workflows.
I feel this as simpler but, If we need the mirror to  remain a full upstream mirror (e.g., for cherry-picks beyond 23.1.0), we can move the stop logic to  automerge at that point.

remaining 2 points I will check it one more time and update.

.github/workflows/sync_from_upstream.yml
```diff
@@ -30,10 +33,25 @@ jobs:
...
-        run: git push origin ${{ matrix.branch }}
+        run: |
+          STOP_TAG="${{ matrix.stop_at_tag }}"
+          if [ -n "$STOP_TAG" ] && git rev-parse "$STOP_TAG" > /dev/null 2>&1; then
+            STOP_COMMIT=$(git rev-parse "${STOP_TAG}^{}")
+            if [ "$(git rev-parse HEAD)" = "$STOP_COMMIT" ]; then
+              echo "Already at $STOP_TAG, nothing to sync."
+              exit 0
+            fi
+            git push origin "${STOP_COMMIT}:refs/heads/${{ matrix.branch }}"
```
navaneethshan, line 54: This will rewind the commits if the TAG appears on an older commit.

.github/workflows/sync_from_upstream.yml
```diff
@@ -30,10 +33,25 @@ jobs:
...
+          fi
       - name: Configure Git Identity
         run: |
           git config --local user.name "github-actions[bot]"
           git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
       - name: Push Changes
-        run: git push origin ${{ matrix.branch }}
+        run: |
+          STOP_TAG="${{ matrix.stop_at_tag }}"
+          if [ -n "$STOP_TAG" ] && git rev-parse "$STOP_TAG" > /dev/null 2>&1; then
```
navaneethshan, line 48: This will check all the branches, not just the specific branch. Good catch Jonathan.

.github/workflows/sync_from_upstream.yml
jonathonpenix: Wouldn't there have to be a `--force` for that to be the case?

Either way, having that be the behavior seems bad--we shouldn't be rewriting history of our public branches.

.github/workflows/sync_from_upstream.yml
```diff
@@ -27,13 +30,34 @@ jobs:
...
+             git merge-base --is-ancestor "$(git rev-parse "${STOP_TAG}^{}")" HEAD; then
+            STOP_COMMIT=$(git rev-parse "${STOP_TAG}^{}")
+
+            if [ "$(git rev-parse HEAD)" = "$STOP_COMMIT" ]; then
+              echo "Already at $STOP_TAG ($STOP_COMMIT), nothing to sync."
+              exit 0
+            fi
+
+            echo "Freezing ${{ matrix.branch }} at $STOP_TAG ($STOP_COMMIT)."
+            git push --force-with-lease origin "${STOP_COMMIT}:refs/heads/${{ matrix.branch }}"
```
jonathonpenix, line 60: Why is `--force-with-lease` used here?

If this is to rollback commits or something, that seems like a bad idea to be doing to public branches

(And, can our automerge even cope with that? For a situation where the force is needed, wouldn't we also have to rollback a bunch of commits in our qualcomm-software-ish branches?)

.github/workflows/sync_from_upstream.yml
```diff
@@ -27,13 +30,34 @@ jobs:
         with:
           ref: ${{ matrix.branch }}
           token: ${{ steps.generate-token.outputs.token }}
+          fetch-depth: 0
```
jonathonpenix, line 33: Why's this needed?

.github/workflows/sync_from_upstream.yml
```diff
@@ -27,13 +30,34 @@ jobs:
...
+        run: |
+          STOP_TAG="${{ matrix.stop_at_tag }}"
+          if [ -n "$STOP_TAG" ] && \
+             git rev-parse "$STOP_TAG" > /dev/null 2>&1 && \
+             git merge-base --is-ancestor "$(git rev-parse "${STOP_TAG}^{}")" HEAD; then
+            STOP_COMMIT=$(git rev-parse "${STOP_TAG}^{}")
+
+            if [ "$(git rev-parse HEAD)" = "$STOP_COMMIT" ]; then
+              echo "Already at $STOP_TAG ($STOP_COMMIT), nothing to sync."
+              exit 0
```
jonathonpenix, line 56: This doesn't seem right to me

HEAD might be at the right spot, but we still might have to push to the remote, no?

PR #595, opened 2026-08-12
[cpullvm] Move, increase RAM for armv7m variants

jonathonpenix:
The current RAM addresses we use for all armv7m variants point to relatively limited regions of RAM (somewhere between 4-8MB depending on how the appropriate Application Notes/QEMU memory setup should be interpreted).

We've seen a few different issues from this:
* We're xfail'ing a few tests due to insufficient RAM.
* When trying to add picolibc v1.8.12 support, no-flash tests are timing out.
    * This _seems_ to be a result of QEMU having 4MB of RAM in these regions and mirroring it with another 4MB. So, zero-ing out 8MB in these tests _also_ zeroes out the .text section as these tests place .text/.data/.bss/etc. in the same segment).

IIUC from the application notes, both the machines we're using for the armv7m variants (mps2-an385 and mps2-an500) have larger blocks of RAM (16MB) allocated at different addresses that don't have the same quirks.

So I think we can point to this area instead to:
1. Allow us to allocate/use more RAM (and thus remove the xfail'ed tests due to RAM limits)
2. Seemingly sidestep the issues observed with the picolibc v1.8.12 no-flash tests.
 
On actual hardware I think this is a _different_ type of RAM, but for QEMU I don't think it matters.

Note that we still don't have identical RAM layouts to ATfE, but this is closer I think.

Note that the "sort-heap-complexity-armv7m" don't describe RAM issues, but these tests are XPASS-ing with this change. Given that they exit with SIGABRT if only given 4MB of RAM and pass if given 6MB of RAM, it seems more likely to me that this was really a issue with the QEMU RAM mirroring rather than what is described in the xfail.

### DIFF
--- a/qualcomm-software/embedded-multilib/json/variants/armv7m_hard_fpv5_d16_nopic.json
+++ b/qualcomm-software/embedded-multilib/json/variants/armv7m_hard_fpv5_d16_nopic.json
@@ -11,8 +11,8 @@
             "QEMU_CPU": "cortex-m7",
             "FLASH_ADDRESS": "0x00000000",
             "FLASH_SIZE": "0x00400000",
-            "RAM_ADDRESS": "0x20000000",
-            "RAM_SIZE": "0x00800000",
+            "RAM_ADDRESS": "0x60000000",
+            "RAM_SIZE": "0x00a00000",
             "LIBRARY_BUILD_TYPE": "minsizerelease"
         },
         "picolibc": {
--- a/qualcomm-software/embedded-multilib/json/variants/armv7m_soft_nofp.json
+++ b/qualcomm-software/embedded-multilib/json/variants/armv7m_soft_nofp.json
@@ -11,8 +11,8 @@
             "QEMU_CPU": "cortex-m3",
             "FLASH_ADDRESS": "0x00000000",
             "FLASH_SIZE": "0x00400000",
-            "RAM_ADDRESS": "0x20000000",
-            "RAM_SIZE": "0x00800000",
+            "RAM_ADDRESS": "0x21000000",
+            "RAM_SIZE": "0x00a00000",
             "LIBRARY_BUILD_TYPE": "minsizerelease"
         },
         "picolibc": {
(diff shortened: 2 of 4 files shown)


jonathonpenix:
Example no-flash picolibc timeouts: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/31636360471/job/94247545438?pr=581

jonathonpenix:
I think the current [failures ](https://github.com/qualcomm/cpullvm-toolchain/actions/runs/31637339664/job/94250869166?pr=595) about test_macros.h are unrelated to this change--they also appear in ex: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/31661537601/job/94327189600#step:6:9395 and/or https://github.com/qualcomm/cpullvm-toolchain/actions/runs/31636360471/job/94247545604?pr=581 which don't have this change.

jonathonpenix:
@bojle Can you please take a look at this and let me know if you have thoughts on this?

It isn't ready to be merged as-is and I don't really understand the issues I'm seeing in picolibc v1.8.12 yet, but I _think_ it intersects with what you described in https://github.com/qualcomm/cpullvm-toolchain/pull/586

jonathonpenix:
Link to complexity.pass.cpp and ranges_sort_heap.pass.cpp unexpectedly passing with this change: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/31744898488/job/94596958449?pr=595

These fail with SIGABRT with 4MB of RAM, so I'm assuming this was another issue with how the memory regions are setup in QEMU rather than what is described in the xfail.

jonathonpenix:
Relevant libc++ tests passing with these changes (changes in test.sh): https://github.com/qualcomm/cpullvm-toolchain/actions/runs/31758406207/job/94639220109?pr=595

bojle:
Hi @jonathonpenix sorry for the delayed reply. This looks good to me. I was unsure how rigorously we have to adhere to the rest of the values in the variants json file given these variants probably come from real customer requirements. If moving the start address to a different section is fine, this is good to go.

bojle (approved):

jonathonpenix:
> I was unsure how rigorously we have to adhere to the rest of the values in the variants json file given these variants probably come from real customer requirements.

These addresses are _broadly_ only for internal testing.

They "leak" out a little bit in the sense that they appear as defaults in picolibc.ld/picolibcpp.ld, but a) these linker scripts aren't really expected to work for "real" images as-is and b) there's overrides present that should be used if someone is trying to use them for "real" (see the comment in picolibc.ld "These values should...").

So I think these we can change freely.

Things that impact the actual build of the library (`COMPILE_FLAGS`, exceptions/rtti/tls model, etc.) we need to be a bit more careful about.

apazos (approved):

PR #596, opened 2026-08-13
[libunwind] Add libcxx/test_support to include flags

bojle:
This PR https://github.com/llvm/llvm-project/pull/214820 adds the use of test_macros.h (present in libcxx/test_support) to the lit feature detection logic. This patch adds the include so that builds can discover this header.


jonathonpenix (approved):
Thanks for tracking this down!

apazos:
Thanks for the quick fix, @bojle!

PR #599, opened 2026-08-15
release/qualcomm-software/23.x: [cpullvm][Github] Use latest successful Linux x86 build in Zephyr tests (#567)

cpullvmbot:
Backport e4328a9

Requested by: @jonathonpenix

### DIFF
--- a/.github/actions/fetch-linux-toolchain/action.yml
+++ b/.github/actions/fetch-linux-toolchain/action.yml
@@ -30,6 +30,18 @@ inputs:
     description: Directory to extract the toolchain into.
     required: false
     default: /opt/cpullvm
+  job-name:
+    description: >
+      Name of the nightly.yml job that must have succeeded for a run to be selected.
+      Only used when run-id is not specified.
+    required: false
+    default: 'build.sh on Linux-x86_64'
+  max-age-days:
+    description: >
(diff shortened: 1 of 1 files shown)


pranav4330:
> Ana Pazos (@apazos) pranav4330 What do you think about merging this PR to the release branch?

I'm okay with merging this we need it to compare and test my Test Report PR.

In the future, we can also create a separate action.yml for Windows. I think it's better to keep the Linux and Windows toolchain acquisition scripts separate.

This way, if one implementation fails, we'll still have the other platform (Linux or Windows) available and functional. It also keeps the platform-specific logic isolated, making maintenance and troubleshooting easier.

pranav4330 (approved):
Looks good to me. Let's merge this as-is, and we can follow up later with a separate Windows action.yml and toolchain acquisition flow if needed.

jonathonpenix:
In hindsight/looking at the failures in https://github.com/qualcomm/cpullvm-toolchain/pull/600, this might not actually be that helpful in the branch since it's using `schedule` as a heuristic to find our "real" nightly builds. Which, doesn't work for our branch builds.

We either need more logic to differentiate the "real" branch nightly builds or the downstream Zephyr scheduler would have to select a runid I think

apazos (commented):
No need to bring it to branch 23.x.

PR #602, opened 2026-08-17
release/qualcomm-software/23.x: [libunwind] Add libcxx/test_support to include flags (#596)

cpullvmbot:
Backport 294e097554afc3f9f62a8f67b2bd998b1e4df155

Requested by: @bojle


jonathonpenix:
Is the corresponding upstream libunwind change that makes this required present in the branch? My understanding is that it isn't.

If not, not sure we should worry about backporting this.

apazos (commented):
This LLVM change is not in the branch yet, so we can skip this for now.

PR #469, opened 2026-07-03
[CPULLVM] - Prepare for LIBCXX tests enablement RISCV.

pawan-nirpal-031:
Manage fixes for RISC-V LIBCXX tests enablement. 

We add xfails for the following variants, but do not enable all of them due to CI timeout limit.
- riscv32gc_ilp32d
- riscv32imac_ilp32
- riscv32imac_ilp32_nopic
- riscv32imac_zba_zbb_ilp32
- riscv32imac_zba_zbb_ilp32_nopic
- riscv32imac_zcb_zcmp_ilp32_nopic
- riscv32imac_zcb_zcmp_zba_zbb_ilp32_nopic
- riscv32imafc_ilp32f
- riscv32imafc_zba_zbb_ilp32f
- riscv32imafc_zcb_zcmp_zba_zbb_ilp32f
- riscv64gc_lp64_nopic
- riscv64gc_lp64d_nopic
- riscv64gc_zba_zbb_lp64_nopic
- riscv64gc_zba_zbb_lp64d_nopic
- riscv64imac_lp64_nopic

We only enable the following riscv variants.
- riscv32gc_ilp32d
- riscv32imac_ilp32
- riscv32imac_zba_zbb_ilp32_nopic
- riscv32imafc_zcb_zcmp_zba_zbb_ilp32f
- riscv64gc_lp64_nopic
- riscv64gc_zba_zbb_lp64d_nopic
- riscv64imac_lp64_nopic

PR #620, opened 2026-08-22
[cpullvm] Add an xqci variant as part of libc++ tests

jonathonpenix:
Previously, xqci variants were omitted as we saw huge numbers of libc++ test
failures in variants without 'a' (see https://github.com/qualcomm/cpullvm-toolchain/pull/469)
and at the time we didn't have any xqci variants with 'a'.

We do now (and these variants turn on C++ libs and `ENABLE_LIBCXX_TESTS`) so
add one such variant in the scheduled libc++ tests--it should be clean.

### DIFF
--- a/.github/workflows/libcxx-test.yml
+++ b/.github/workflows/libcxx-test.yml
@@ -7,8 +7,6 @@
 # Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries. 
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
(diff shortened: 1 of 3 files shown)


jonathonpenix:
Successful run of c++ tests: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/32618268207/job/97142376093?pr=620

apazos (approved):
Thanks @jonathonpenix for enabling this config now that there is a variant with atomics enabled.

PR #619, opened 2026-08-22
[cpullvm][Multilib] Add aligned AArch64 variants

jonathonpenix:
Add aligned alternatives of the following variants:
- aarch64a_tlsie
- aarch64a_pacret_bkey_bti_tlsie
- aarch64a_soft_nofp_tlsie
- aarch64a_soft_nofp_pacret_bti

So far these are the only variants where folks have needed unaligned variants, so we're just going to duplicate these for now.

See https://github.com/qualcomm/cpullvm-toolchain/issues/388 for additional context around our Arm/AArch64 variant alignment situation.

This is a port of #448 from 22.x

PR #623, opened 2026-08-24
release/qualcomm-software/23.x: [cpullvm][Multilib] Add aligned AArch64 variants (#619)

cpullvmbot:
Backport 2ecb964

Requested by: @jonathonpenix

### DIFF
--- a/qualcomm-software/embedded-multilib/json/multilib.json
+++ b/qualcomm-software/embedded-multilib/json/multilib.json
@@ -13,6 +13,12 @@
             "libraries_supported": "picolibc"
         },
         {
+            "variant": "aarch64a_aligned_tlsie",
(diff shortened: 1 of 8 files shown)


pzhengqc:
> Ana Pazos (@apazos) pzhengqc What do you think about merging this PR to the release branch?

I think it makes sense to merge this PR into the release branch. Not sure if any team requested these variants for this release though. If there's any concern about the timing, we could also do a patch release later maybe?

jonathonpenix:
Yeah, this came up with a customer image where they use/need -mno-unaligned-access so they don't have appropriate multilibs right now (which currently manifests as multilib not found warnings/errors when building as long as we want to also have a unaligned variant + require strict matching)

So I'm hoping we can sneak this in before we freeze later this week

apazos (approved):

PR #625, opened 2026-08-25
[cpullvm] Unify Linux AArch64 Host Compiler to Clang 22.1.8 on GitHub Ubuntu Runner

pranav4330:
Align the Linux AArch64 GitHub Ubuntu runner with the host compiler version used across other build environments.

This change installs and uses Clang 22.1.8 on the GitHub-hosted AArch64 Ubuntu runner, ensuring consistent compiler versions across all supported build configurations and reducing toolchain-related variations.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -16,6 +16,7 @@
 #
 # Other build configurations such as building the runtimes on non-Linux
 # x86_64 platforms are built and tested separately.
+# clang-22 for linux aarch64 github runner.(testing)
 
 name: Nightly Build and Test
 
--- a/qualcomm-software/scripts/install_dependencies.sh
+++ b/qualcomm-software/scripts/install_dependencies.sh
@@ -1,14 +1,24 @@
 # Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+CLANG_VERSION = 22
+
 sudo apt-get update
+
+# Add LLVM APT repo (required for clang-22)
+wget -qO llvm.sh https://apt.llvm.org/llvm.sh
+chmod +x llvm.sh
+sudo ./llvm.sh ${CLANG_VERSION}
+
 # Install swig and libedit-dev used by lldb and
 # libc++-dev required for eld tests
-sudo apt-get install -y swig libedit-dev clang-19 libc++-19-dev
+sudo apt-get install -y swig libedit-dev libc++-${CLANG_VERSION}-dev
 
-# Set default clang version to clang-19
-sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-19 100
-sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-19 100
+# Set default clang version to clang-22
+sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-${CLANG_VERSION} 100
+sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-${CLANG_VERSION} 100
+sudo update-alternatives --set clang /usr/bin/clang-${CLANG_VERSION}
+sudo update-alternatives --set clang++ /usr/bin/clang++-${CLANG_VERSION}
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
 pip install meson==1.10.0


.github/workflows/nightly.yml
```diff
@@ -16,6 +16,7 @@
 #
 # Other build configurations such as building the runtimes on non-Linux
 # x86_64 platforms are built and tested separately.
+# clang-22 for linux aarch64 github runner.
```
jonathonpenix: If this is just a temporary change to trigger the workflow, that is fine--please just make sure it is removed once the build completes/before merging.

Otherwise, please remove this--it isn't providing any particularly helpful information and keeping this updated is just going to be noise.

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -1,14 +1,27 @@
...
+
+echo "deb [signed-by=/usr/share/keyrings/llvm-archive-keyring.gpg] \
+http://apt.llvm.org/$(lsb_release -sc)/ llvm-toolchain-$(lsb_release -sc)-22 main" \
+    | sudo tee /etc/apt/sources.list.d/llvm.list
+
 sudo apt-get update
 # Install swig and libedit-dev used by lldb and
 # libc++-dev required for eld tests
-sudo apt-get install -y swig libedit-dev clang-19 libc++-19-dev
+sudo apt-get install -y swig libedit-dev clang-22 libc++-22-dev
```
jonathonpenix:
```suggestion
sudo apt-get install -y swig libedit-dev clang-22 libc++-22-dev
```

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -1,14 +1,27 @@
...
 sudo apt-get update
 # Install swig and libedit-dev used by lldb and
 # libc++-dev required for eld tests
-sudo apt-get install -y swig libedit-dev clang-19 libc++-19-dev
+sudo apt-get install -y swig libedit-dev clang-22 libc++-22-dev 
 
 # Set default clang version to clang-19
-sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-19 100
-sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-19 100
+sudo update-alternatives --install /usr/bin/clang   clang   /usr/bin/clang-22 100
```
jonathonpenix: Please don't add the extra whitespace

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -1,14 +1,27 @@
...
+
+sudo apt-get update
+
+# Add LLVM APT repo (required for clang-22)
+wget -qO - https://apt.llvm.org/llvm-snapshot.gpg.key \
+    | sudo gpg --dearmor -o /usr/share/keyrings/llvm-archive-keyring.gpg
+
+echo "deb [signed-by=/usr/share/keyrings/llvm-archive-keyring.gpg] \
+http://apt.llvm.org/$(lsb_release -sc)/ llvm-toolchain-$(lsb_release -sc)-22 main" \
+    | sudo tee /etc/apt/sources.list.d/llvm.list
```
jonathonpenix: Is there a reason not to just use the installation script LLVM provides? https://apt.llvm.org/#llvmsh

qualcomm-software/scripts/install_dependencies.sh
jonathonpenix: I wonder if we're missing lld here--see the /usr/bin/ld warnings in the AArch64 build

This _might_ be fixed by using the install script as I think it'll pull everything if not otherwise specified.

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -1,14 +1,27 @@
...
-sudo apt-get install -y swig libedit-dev clang-19 libc++-19-dev
+sudo apt-get install -y swig libedit-dev clang-22 libc++-22-dev 
 
 # Set default clang version to clang-19
-sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-19 100
-sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-19 100
+sudo update-alternatives --install /usr/bin/clang   clang   /usr/bin/clang-22 100
+sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-22 100
+sudo update-alternatives --set clang   /usr/bin/clang-22
+sudo update-alternatives --set clang++ /usr/bin/clang++-22
```
jonathonpenix: Yikes, it looks like our old builds weren't actually using clang-19 for the main build?

jonathonpenix (commented):
Re: the PR description/title:
- Please don't mark this as `[clang]`, it isn't a change in clang
- It seems a bit misleading to say that this is aligning with other build environments seeing as no other builds use 22 right now

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -1,14 +1,27 @@
...
 # libc++-dev required for eld tests
-sudo apt-get install -y swig libedit-dev clang-19 libc++-19-dev
+sudo apt-get install -y swig libedit-dev clang-22 libc++-22-dev 
 
 # Set default clang version to clang-19
-sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-19 100
-sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-19 100
+sudo update-alternatives --install /usr/bin/clang   clang   /usr/bin/clang-22 100
+sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-22 100
+sudo update-alternatives --set clang   /usr/bin/clang-22
```
jonathonpenix:
```suggestion
sudo update-alternatives --set clang /usr/bin/clang-22
```

.github/workflows/nightly.yml
pranav4330: yes its a temporary change

qualcomm-software/scripts/install_dependencies.sh
pranav4330: llvm.sh is a convenience script that installs the entire LLVM toolchain. we don't need all of it, just clang-22 and libc++. The manual APT approach installs only what we need.

qualcomm-software/scripts/install_dependencies.sh
pranav4330: I think default ld is being picked up so i will add lld-22

qualcomm-software/scripts/install_dependencies.sh
jonathonpenix: In general I think we do want/need an entire LLVM toolchain.

The "extra" we'd get (even from doing an `all`) out of the LLVM script is really not much more than what we're already getting once you account for all the packages we're already transitively pulling in (and, we won't miss dependencies that we want like lld, etc.).

If it really consumes _that_ much more space/time, then maybe, but I'm skeptical that is the case.

The other side too is down the road I think it is far clearer for people to see what is going on via LLVM's script (which, they'll maintain as there's updates, etc.) rather than doing our own thing (which, AFAICT is already different than what LLVM documents).

pranav4330: I have addressed this in new push and used script by LLVM

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -1,14 +1,24 @@
 # Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+
```
sriyalamar, line 5: Shall we define CLANG_VERSION=22 and refer to it as ${CLANG_VERSION}?

pranav4330: I think its not necessary, as this is working as expected.

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -1,14 +1,24 @@
...
+sudo apt-get update
```
jonathonpenix, line 6: Do we actually need to be doing this twice? Once here and once below

Please also remove the extra newline

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -1,14 +1,24 @@
...
+# Add LLVM APT repo (required for clang-22)
+wget -qO llvm.sh https://apt.llvm.org/llvm.sh
+chmod +x llvm.sh
+sudo ./llvm.sh 22
+
 sudo apt-get update
 # Install swig and libedit-dev used by lldb and
 # libc++-dev required for eld tests
-sudo apt-get install -y swig libedit-dev clang-19 libc++-19-dev
+sudo apt-get install -y swig libedit-dev clang-22 libc++-22-dev
```
jonathonpenix: Please remove clang-22--it is already installed in the script above.

For libc++-22-dev, it is still needed--I'm fine if we leave this or add `all` to the script invocation above.

qualcomm-software/scripts/install_dependencies.sh
jonathonpenix: I think it is a good suggestion--next time we update versions we only need to update one spot.

pranav4330:
hello @jonathonpenix  i have made the changes can you check if i missed something, also @navaneethshan can you look at this as well?

qualcomm-software/scripts/install_dependencies.sh
pranav4330: thanks for suggesting @sriyalamar I have made that change.

qualcomm-software/scripts/install_dependencies.sh
```diff
@@ -1,14 +1,24 @@
...
-sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-19 100
+# Set default clang version to clang-22
+sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-${CLANG_VERSION} 100
+sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-${CLANG_VERSION} 100
+sudo update-alternatives --set clang /usr/bin/clang-${CLANG_VERSION}
+sudo update-alternatives --set clang++ /usr/bin/clang++-${CLANG_VERSION}
 
 # Install meson. eld support was added in v1.9.0, so we need at least that.
-pip install meson==1.10.0
+pip install meson==1.10.0
```
jonathonpenix, line 24: It looks like something weird is going on with whitespace--please remove this change

(probably some line ending or invisible characters?)

jonathonpenix (commented):
~One last nit, but otherwise this should be close.~

~When fixing whatever is changing in the meson line, I think you can remove the test change in nightly.yml to prepare for merging.~

Added a nit, but I think it is also a bit concerning we're still seeing the /usr/bin/ld warnings--seems like something still is off.

qualcomm-software/scripts/install_dependencies.sh
jonathonpenix: Hm, still seeing the /usr/bin/ld warnings

PR #630, opened 2026-08-26
[cpullvm] Add release/23.x to automerge for repo version updates

sriyalamar:
Add release/23.x -> release/qualcomm-software/23.x entry back to the automerge workflow so that repo version updates are triggered for the 23.x release branch.


jonathonpenix (approved):
Thanks!

And thanks for catching that we disabled the repo version updates as well--I missed that earlier.

I think this makes sense to me--we have _syncing from LLVM_ disabled in the branch, so we should be safe on that front hopefully.

PR #636, opened 2026-08-31
[cpullvm] Stop automerge for release/23.x (code freeze)

sriyalamar:
Remove release/23.x from the automerge workflow matrix to stop automatic updates in version.json in the release/qualcomm-software/23.x.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -19,8 +19,6 @@ jobs:
         branches:
           - from_branch: main
             to_branch: qualcomm-software
-          - from_branch: release/23.x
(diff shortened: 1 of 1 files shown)


jonathonpenix:
I think when we go to do our final code freeze we might want to also strip out the `trackingBranch`s from `versions.json`.

After a certain point, updates should be done manually and I think it maybe makes it a bit more clear that there's no more auto-updating these in the branch.

If we expect we might want to turn on automerging again in the branch, I'm fine if we omit this for now, but I think it is worth doing before we finalize the release.

apazos (approved):

PR #637, opened 2026-08-31
23.x: [cpullvm] Remove `trackingBranch`es from versions.json

jonathonpenix:
We've frozen so all updates should happen manually. This should also make it more clear that automerge is off/shouldn't be re-enabled.


jonathonpenix:
This was omitted from https://github.com/qualcomm/cpullvm-toolchain/pull/636, so adding this here.

apazos (approved):

sriyalamar:
LGTM, Thanks @jonathonpenix

sriyalamar (approved):

PR #640, opened 2026-09-01
[cpullvm] Update workflow to include Ubuntu version tag in package name

sriyalamar:
Add Ubuntu version information to the package name generated by the workflow for easier identification and tracking.


jonathonpenix (changes requested):
Please check #638--I think we want to move towards always appending the Ubuntu version (across all packages, and not just in our branches).

PR #644, opened 2026-09-01
23.x: [cpullvm] Add Ubuntu major version information to package name by default

jonathonpenix:
Old example: cpullvm-24.0.0-Linux-x86_64.tar.xz
New example: cpullvm-24.0.0-Linux-x86_64-Ubuntu24.tar.xz

Non-Ubuntu OS'es should be unaffected.

Note that this changes the final package name only--the "wrapper" packages (ex: cpullvm-packages-build.sh-Linux-x86) are left unchanged. In the past we added the separate Ubuntu versions in different workflows, so things can be distinguished easily still. Maybe this will need to change, but for now just follow what we've done in the past.

Assisted-by: claude

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -155,7 +155,7 @@ jobs:
           path: |
             build*/cpullvm-*.tar.xz
             build*/cpullvm-*.zip
-            !build/cpullvm-*-Linux-x86_64.tar.xz
+            !build/cpullvm-*-Linux-x86_64*.tar.xz
           retention-days: 7
 
       - name: Test
--- a/qualcomm-software/CMakeLists.txt
+++ b/qualcomm-software/CMakeLists.txt
@@ -461,7 +461,18 @@ endif()
 if (PACKAGE_SYSTEM_NAME)
     set(CPACK_SYSTEM_NAME "${PACKAGE_SYSTEM_NAME}-${processor_name}")
 else()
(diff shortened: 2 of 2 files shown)


jonathonpenix:
Manual cherry-pick of https://github.com/qualcomm/cpullvm-toolchain/pull/638 to work around eld test failures

jonathonpenix:
Cancelling the nightly release workflow run to hopefully free up some resources for other builds. It should work identically to nightly.yml here.

Yethendra (approved):
LGTM

sriyalamar:
Hi @jonathonpenix, Currently, the package name is: cpullvm-packages-build.sh-Linux-x86
and the internal tar file is: cpullvm-23.1.0-Linux-x86_64-Ubuntu24.tar.xz

If we do not plan to include the Ubuntu version in the package name (for example, cpullvm-packages-build.sh-Linux-x86-Ubuntu24), then the PR LGTM. Thanks!

jonathonpenix:
Yeah, the "external" package names are left alone for now.

Maybe down the road we'll have to change/rethink this, but if we follow what we did in the past for ex: Ubuntu 22 and 24 builds in 22.x and build the extra OS as a separate workflow, I think it is ok to leave the package names as-is. 

Changing the package names also means we have to be careful to update other workflows (like Zephyr tests, etc.) that are sensitive to these names so it seems like more hassle than it is worth right now.

apazos (approved):

PR #657, opened 2026-09-08
23.x: [cpullvm] Include lldbPluginScriptInterpreterPython in distribution

jonathonpenix:
Otherwise it is excluded and we have broken symlinks in our final install/distribution (from _lldb.abi3.so to the missing liblldbPluginScriptInterpreterPython.so).


jonathonpenix:
(this would/will need to go in mainline as well, but starting here)

efriedma-quic (approved):
LGTM

apazos (approved):

PR #660, opened 2026-09-08
[cpullvm] Add xqci variants with exceptions, rtti enabled

jonathonpenix:
Two new variants were added, both built with exception and rtti support
- riscv32ima_xqci_ilp32_exn_rtti_nopic
- riscv32ima_zinx_xqci_ilp32_exn_rtti_nopic

Note that these will now be matched rather than the old riscv32ima_xqci_ilp32_nopic or riscv32ima_zinx_xqci_ilp32_nopic variants unless users explicitly pass -fno-exceptions AND -fno-rtti, even when building ex: C code. If they pass both, they should get the old variants.

Note also that riscv32ima_xqci_ilp32_exn_rtti_nopic is being added for libc++ testing--I think it is worth having at least one extension + rtti variant enabled since this is new for us.


jonathonpenix:
Sorry, marking this as a draft since I have to go back and add a bunch of xfails.

jonathonpenix:
Right now there are ~4 libc++ failures that need to be sorted out with these variants:

- libcxx/utilities/exception_guard.odr.sh.cpp
    - Link flags in the test just need to be re-arranged, this is a change to go to upstream LLVM
- std/re/re.alg/re.alg.match/exponential.pass.cpp and std/re/re.alg/re.alg.search/exponential.pass.cpp
    - Not sure what the issue is. Haven't looked into it
- std/strings/basic.string/string.modifiers/string_replace/size_size_string_size_size.pass.cpp
    - Fails to link with memory region overflows. Not sure if there's something going wrong here or if it really just does end up being too large. Need to look more into it.

Otherwise, other tests (picolibc, compiler-rt, libunwind, libcxxabi, multilib) are passing locally

pgodeq:
Thanks Jonathon for putting this together.  I had missed adding -fno-exceptions and -fno-rtti for the earlier variants.
These changes make sense.

pgodeq:
Hi Jonathon, should we add '-fno-exceptions -fno-rtti' to existing variants 
- riscv32ima_xqci_ilp32_nopic.json
- riscv32ima_zinx_xqci_ilp32_nopic.json

I see that we have mentioned these flags in multilib.json for existing variants so thought of checking.

jonathonpenix:
Maybe? I imagine it wouldn't hurt anything. Not sure if it will change anything though--I think libcxx/libcxxabi/libunwind all will have them set appropriately and I'm not sure it should matter for other projects.

But we've made changes like that to be explicit about flags in the past, so maybe it'd be worth doing anyway. But if we do it for these variants, I guess I'd prefer we do it for all of our variants built without exception support.

pgodeq:
> But we've made changes like that to be explicit about flags in the past, so maybe it'd be worth doing anyway. But if we do it for these variants, I guess I'd prefer we do it for all of our variants built without exception support.

I agree that these flags should be explicit to avoid confusion. I have observed few cases where customers add -fno-exception explicitly.

qualcomm-software/scripts/test.sh
```diff
@@ -25,3 +25,5 @@ ninja check-all
 ninja check-llvm-toolchain
 ninja check-cxxabi
 ninja check-unwind
+ninja check-cxx-riscv32ima_xqci_ilp32_exn_rtti_nopic
+ninja check-cxx-riscv32ima_zinx_xqci_ilp32_exn_rtti_nopic
```
jonathonpenix: These are for testing only, will remove them once the build finishes.

jonathonpenix:
> > But we've made changes like that to be explicit about flags in the past, so maybe it'd be worth doing anyway. But if we do it for these variants, I guess I'd prefer we do it for all of our variants built without exception support.
> 
> I agree that these flags should be explicit to avoid confusion. I have observed few cases where customers add -fno-exception explicitly.

Sorry, re-reading this I'm confusing myself so just to clarify, you're asking about adding -fno-exceptions for these variants in their variant-specific .json build flags, not multilib.json right?

variant-specific .json build flags seems fine to me to add -fno-exceptions for these variants (and all our other variants without exception support)--should be NFC I think, just more explicit like mentioned above.

Adding -fno-exceptions to multilib.json for all the other variants without exception support is probably where we should get to eventually, but for now we also do have a lot of customers that *don't* explicitly specify -fno-exceptions (especially if their project is just C code, etc.) which is why we originally didn't require that flag for multilib matching. Which, again, we might want to do/revisit at some point, but I think that is a slightly bigger change that we'd want to be ready to help users through.

jonathonpenix:
Precheckin with everything passing (including the libc++ tests): https://github.com/qualcomm/cpullvm-toolchain/actions/runs/35686876355/job/106615477269?pr=660

jonathonpenix:
I think this should be more or less ready for review.

I'll upstream push the patchfile as a PR against upstream in a bit (with some rewording, but should be the same change).

I think I'd like to do the "add -fno-exceptions/rtti to build flags" as a follow-up if folks are ok with it (and do it for either all the variants or no variants in mainline). There's a few things I want to check on (how it interacts with the non-libc++ runtime builds) before going forward with that. Experimenting with it here: https://github.com/qualcomm/cpullvm-toolchain/pull/681

jonathonpenix:
> I think I'd like to do the "add -fno-exceptions/rtti to build flags" as a follow-up if folks are ok with it (and do it for either all the variants or no variants in mainline). There's a few things I want to check on (how it interacts with the non-libc++ runtime builds) before going forward with that. Experimenting with it here: #681

@pgodeq I think I'm going to abandon this/not add the -fno-exceptions/-fno-rtti to build flags--see https://github.com/qualcomm/cpullvm-toolchain/pull/681#issuecomment-5802036307. I guess if we notice specific issues from _not_ doing it maybe we can revisit it.

PR #661, opened 2026-09-08
[cpullvm] xfail libunwind test dwarf_expression_stack.pass.cpp

jonathonpenix:
This explicitly uses fork() and waitpid() which I don't think are supported by picolibc for our embedded targets. So just xfail the test.

Right now this only impacts our AArch64 variants, but just xfail it for everything since all of our Arm/AArch64/RISC-V variants would have the same issue.

### DIFF
--- a/qualcomm-software/embedded-runtimes/test-support/xfails.py
+++ b/qualcomm-software/embedded-runtimes/test-support/xfails.py
@@ -461,6 +461,16 @@ def main():
                 "installed in the sysroot. ATfE does not encounter this failure because "
                 "their aarch64a_soft_nofp variant has ENABLE_CXX_LIBS=ON.",
         ),
+        XFail(
+            name="libunwind dwarf_expression_stack.pass.cpp",
+            testnames=[
+                "dwarf_expression_stack.pass.cpp",
+            ],
+            result=NewResult.XFAILED,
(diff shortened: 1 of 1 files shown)


apazos (approved):

efriedma-quic (commented):
I'm okay with this, but we should open a bug upstream.  (It looks like this test was added in https://github.com/llvm/llvm-project/pull/220662 ?)

jonathonpenix:
I'm not opposed to filing something upstream, but I guess what are we hoping to get at with the bug? Are we looking for a better way to keep these sorts of tests (that use fork(), etc.) isolated so we don't have to xfail them? Or something else?

As far as I can tell there's nothing really going wrong here, it's just that we're running the test in an environment where fork() etc. isn't (will probably never be?) supported.

efriedma-quic:
If the test is structured in a way that can't work, they can add an appropriate UNSUPPORTED line to the test.

pranav4330 (commented):
Is this good for merge? i can see the linux x86 passing here.

pranav4330 (approved):

apazos (approved):

jonathonpenix:
Will create a follow-up issue in LLVM

PR #663, opened 2026-09-09
[cpullvm] Fix Windows LLDB Python plugin distribution

sriyalamar:
## Summary

Exclude `lldbPluginScriptInterpreterPython` from Windows distribution components because the plugin is not built when dynamic Python script interpreters are disabled on Windows.

### DIFF
--- a/qualcomm-software/CMakeLists.txt
+++ b/qualcomm-software/CMakeLists.txt
@@ -172,7 +172,6 @@ set(LLVM_DISTRIBUTION_COMPONENTS
     liblldb
     lld
     lldb
-    lldbPluginScriptInterpreterPython
     lldb-argdumper
     lldb-dap
     lldb-instr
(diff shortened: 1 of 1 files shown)


qualcomm-software/CMakeLists.txt
```diff
@@ -234,6 +233,9 @@ set(LLVM_DISTRIBUTION_COMPONENTS
     LTO
     CACHE STRING ""
 )
+if(NOT WIN32)
+    list(APPEND LLVM_DISTRIBUTION_COMPONENTS lldbPluginScriptInterpreterPython)
```
jonathonpenix: It might be worth moving this down into the same check as where we enable LLDB_ENABLE_DYNAMIC_SCRIPTINTERPRETERS  since that is what controls whether we need this or not.

It'd also resolve the mix between WIN32 and CMAKE_HOST_WIN32--I'm not actually sure what the right one is here, but IIUC they aren't the same and we need the checks to both fire at the same time.

(In practice they might be essentially the same since we never cross compile the toolchains, but I'm not sure if there are other 'gotchas' I'm missing)

jonathonpenix (commented):
Thanks for catching/fixing this.

I'm ok if this goes in as-is, but if builds haven't finished yet it might be worth resolving WIN32 vs CMAKE_HOST_WIN32 just to avoid any possible confusion in the future. Maybe just putting the new change in the check to set LLDB_ENABLE_DYNAMIC_SCRIPTINTERPRETERS  would be easiest

jonathonpenix:
Just as an example of what I mean, see https://github.com/qualcomm/cpullvm-toolchain/pull/665/changes

I don't mind too much which way we go for the branch (I think they should be the same for what we're doing), just need to pick one.

jonathonpenix (approved):

PR #672, opened 2026-09-12
[cpullvm] Add xfails for riscv32ima_zinx_xqci_ilp32_nopic

jonathonpenix:
Do this so:
- We have a clean baseline in case anyone wants to run libc++ tests for this variant (and it should be clean in general)
- It is helpful to see these for comparison with similar variants (as in ex: https://github.com/qualcomm/cpullvm-toolchain/pull/660)

I'm not going to add it to test_libcxx.sh as we need to limit the number of variants we add there to keep the runtime within reason.


jonathonpenix:
Run of libc++ tests for this variant is clean with these changes: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/34667128736/job/103481202277?pr=672

pgodeq (approved):
Looks good to me.

PR #674, opened 2026-09-15
[DNM][RISCV] Input section aware machine outlining

quic-garvgupt:
This is the first part of the machine outliner fix that stands on its own and is in no way related to the LTO + linker script changes.

I will push the second part as a different commit to this PR which will be dependent on the LTO + linker script changes and should resolve the outlining issue observed when lto is enabled with linker script


quic-garvgupt:
I will push the LTO and linker script-specific changes to this PR as a separate commit.

For now, the first commit contains a generic improvement to the Machine Outliner. This serves as the foundational change for the subsequent LTO and linker script-related enhancements. The change is self-contained and independent of the upcoming LTO and linker script patches.

llvm/include/llvm/CodeGen/TargetInstrInfo.h
```diff
@@ -2434,3 +2434,8 @@ template <> struct DenseMapInfo<TargetInstrInfo::RegSubRegPair> {
 } // end namespace llvm
 
 #endif // LLVM_CODEGEN_TARGETINSTRINFO_H
+  // Return true if the target supports outlining candidates from multiple
+  // input sections in one pass. When enabled, candidates are partitioned by
+  // input section before target-specific outlining checks are performed.
+  virtual bool supportsSectionAwareOutlining() const { return false; }
+@@ -2226,0 +2228,5 @@ public:
```
lenary: Something strange has happened to the diff here. Is this a github bug or is there something strange in the actual code here?

lenary (commented):
This looks broadly right to me.

llvm/include/llvm/CodeGen/TargetInstrInfo.h
quic-garvgupt: I am not sure either how it crept in. I have removed it from the latest patchset

PR #677, opened 2026-09-21
[cpullvm] Remove dwarf_expression_stack.pass.cpp xfail

jonathonpenix:
This was gated behind `has-unix-headers` in [1] upstream to prevent it from running on baremetal targets, so we shouldn't need to xfail it downstream anymore.

[1] https://github.com/llvm/llvm-project/commit/be60b244d1f9fa798c8922e313679035ad9c2ee5


efriedma-quic (approved):
LGTM

